PR #21, opened 2022-12-01
added L<->M conversions

stephenmwilkins:


christopherlovell (changes requested):
Remove global variables, place inside dedicated method

PR #33, opened 2022-12-15
Line grid update

christopherlovell:
Few big changes here:
- removed the open hdf5 files in line reading, as at risk of corrupting files
- merged SpectralGrid and LineGrid - motivation for this is that they share pretty much all the same functionality, reducing code maintenance overhead
- line names now contained in `self.line_list` 
- grids for individual lines are contained in `self.lines`
- new method `fetch_lines` grabs the line grid for a given line from the hdf5 file. This has options to save to the grid object (stored in self.lines as a dictionary with luminosity, continuum and wavelength keys) or just return dynamically. The motivation for this is that users may not wish to load all line grids into memory - could be quite large. This allows users to only load those lines they need, and also load them dynamically if they know they're not going to use them again.
- `get_line_info` is now updated to use `fetch_line`. Also now returns a dictionary with keys.
- updated lines example script

This was the most efficient way I could come up with, happy to take suggestions for improvement. I can't see any breaking changes throughout the code, however this will break dependent codes that used `LineGrid`


stephenmwilkins:
Seems fine, though I haven't yet tested the functionality. Since we have a single Grid now we can probably merge Grid and SpectralGrid right?

christopherlovell:
merged, kept the name as `SpectralGrid` for compatibility reasons, we may want to update that in the future

PR #37, opened 2022-12-15
added Yggdrasil

aswinpvijayan:
@christopherlovell check if works and pep8
I have added the nebular processed ones as well, however not sure if you want to process the raw spectra through cloudy


christopherlovell:
few updates:
- made pep8 compliant
- changed script name and the name of the output grid files (we may want to add other popIII models in future)
- added example to `examples/grids` folder

christopherlovell (approved):

PR #85, opened 2023-01-17
Line of sight

aswinpvijayan:
## Issue Type

- [ ] Bug
- [ ] Document
- [x] Enhancement

## Checklist

Added standard dust attenuation/extinction curve in dust.py. curve.attenuate() gives the transmission


WillJRoper:
One thing that's just dawned on me is we will probably want to revisit the joint configuration here if we have our own threadpooling functionality implemented 😬

WillJRoper:
And I have to be like this... but there's missing doc strings in dust.py and classes are missing the 
```
Methods
---------
```
section describing the methods.

aswinpvijayan:
Its now updated with needed docstrings and example plot

PR #88, opened 2023-01-18
Fix cloudy

christopherlovell:
## Issue Type

- [ ] Bug
- [ ] Document
- [x] Enhancement

## Checklist
- [N.A.] I have added tests that prove my fix is effective or that my feature works
- [N.A.] New and existing unit tests pass locally with my changes


stephenmwilkins:
Looks good. A couple of thoughts:

- probably needs more commenting to make Will happy, but those are probably bits I originally wrote. For example, describing how log10U is calculated.

- add comments to the example param files spelling out what they do.

- the machine specific stuff could be rephrased in terms of the job scheduler (e.g. slurm). There could be a dictionary noting the scheduler for different machines?

WillJRoper:
I left off this because my cloudy understanding is minimal, but yes more comments. 

For the record, it's not "to make Will happy", it's too produce actually releasable code that people can use 🙄 😂

christopherlovell:
1) agreed, but I think with this I'm just consolidating to a single versino of `cloudy.py`. We can tackle commenting later
2) will do that now
3) nice idea, but we might have to account for different versions of those schedulers. I'd argue leave as it is for now, and we can amend later if we get a big demand to run on other machines

christopherlovell:
I didn't write much of the content of `cloudy.py` so I will leave for someone else to comment thoroughly. Up to you whether you want to do this in this PR or later. 

I have already gone through and made it pep8 compliant, and added a few docstrings, which took a big chunk of time. 

The purpose of this PR is to get the cloudy functionality in `generate_grids` working again on the main branch. The only reason for consolidating the `cloudy.py` scripts was to avoid confusion. Tidying them up is left as an exercise for the reader...

stephenmwilkins (approved):

PR #107, opened 2023-01-20
added parametric morphology and some bug fixes

stephenmwilkins:
## Issue Type

- [ ] Bug
- [ ] Document
- [ ] Enhancement

## Checklist


WillJRoper (changes requested):
`morphology.img` should be part of `ParametricImage` not the morphology class, I can modify this though. 

I also would remove the image creation with the morphology object from the initialisation of the image. Note that allowing `morphology=None` as the default argument means initialisation will fail if it is not provided. It should either be required or not done at initialisation. Also, `filters` is a required argument, are you always going to have a filter with a parametric model?

WillJRoper:
Also, you ticked that you have modified the docs which I'm pretty sure is a lie 😂 don't just tick boxes to dodge the check failing 😂

stephenmwilkins:
> `morphology.img` should be part of `ParametricImage` not the morphology class, I can modify this though.
> 
> I also would remove the image creation with the morphology object from the initialisation of the image. Note that allowing `morphology=None` as the default argument means initialisation will fail if it is not provided. It should either be required or not done at initialisation. Also, `filters` is a required argument, are you always going to have a filter with a parametric model?

The reason for morphology.img being there was because I could imagine a morphology function that didn't provide method to call with xx, yy. For example if you wanted a collection of point sources.

WillJRoper (approved):

WillJRoper:
I approved it but the linter caught some issues (similar to my filters.py issues 😅). Rectify those and it'll merge.

WillJRoper:
Still not sure why Morphology needs to have `.img` but I will take a look and see if I can relocate it to allow for that functionality. It really shouldn't have imaging functionality. That should be with the images.

stephenmwilkins:
> Still not sure why Morphology needs to have `.img` but I will take a look and see if I can relocate it to allow for that functionality. It really shouldn't have imaging functionality. That should be with the images.

I can understand that but I think it's neater to include all the different morphological models in one code. Like I said, some morphological models can't be described by a function that you pass xx, yy to.

PR #110, opened 2023-01-23
Refactored filters.py

WillJRoper:
I have refactored filters.py to simplify the front end for the user. Now there is only FilterCollection and Filter with methods for different filter types.

This also includes a bunch of nice to have features:
- Being able to add filter collections together.
- Being able to define a filter explictly by transmission curve and wavelength.
- Combine SVO, top hat and generic filters in a single FilterCollection.
- Use an apply_filter method on a filter object to apply the transmission curve to an arbitrary dimension array as long as the final dimension is wavelength. 

Tested but not extensively for non-SVO filters.

## Issue Type

- [ ] Bug
- [ ] Document
- [x] Enhancement

## Checklist

- [N.A] I have added tests that prove my fix is effective or that my feature works


stephenmwilkins (approved):

stephenmwilkins:
Looks good to me, just need to resolve the conflicts but that should be you just overwriting whatever changed.

PR #125, opened 2023-02-06
Update fsps grid creation

stephenmwilkins:
- stellar grid generation re-written for BC03, BC03-2016, and FSPS. BC03 and BC03-2016 will not work from download because they require the use of a Fortran program to convert some files. :( 
- fixed bugs for creating grids varying cloudy parameters (e.g. U, alpha)
- added scripts for 3D grid generation. Created a low-resolution test 3D grid.
- some tweaks to `Grid` and grid examples.
- started consolidating plots in `plots.py`.


christopherlovell (approved):
There is a lot here, with some changes to core synthesizer methods, in addition to the grid creation. Can't test thoroughly but trust it will be okay

PR #129, opened 2023-02-06
Steve fixes

stephenmwilkins:
Fixed broadband photometry units by changing `lnu` to `_lnu`.
Expanded example.


WillJRoper (approved):
Not sure why you've packaged the example changes in this PR? But will let through.... don't remove the checklist...

PR #136, opened 2023-02-08
fixed particle galaxy intrinsic spectra generation

christopherlovell:
## Issue Type

- [ ] Bug
- [ ] Document
- [ x] Enhancement

## Checklist


WillJRoper (approved):
Looks good to me. Possibly some future iteration needed. One question would be how often we are realistically going to need stellar and intrinsic. If so it might be worth including a `get_all_spectra` variant to avoid the double calculation overhead.

PR #167, opened 2023-05-30
Fix particle seds

christopherlovell:
Fixes the methods in `galaxy/particle.py` to match those in `galaxy/parametric.py` 

Some other cosmetic modifications to plotting methods and to add optional verbosity

## Issue Type

- [ ] Bug
- [ ] Document
- [ x ] Enhancement

## Checklist


WillJRoper (changes requested):
Is this really the best way to do these? It's basically undone all of the simplifications I spent time doing in Malta. We had a long discussion about not having tons of methods for individual spectra types and instead handling this in a single front facing function which takes the `spectra_type` argument. 

Sure this works for now to get plots done but it undoes a bunch of work and recomplicates things. Simplicity is paramount on the front end.

christopherlovell:
As discussed off thread, I think there was a conclusion that a single function became super complicated in reality. But that maybe especially the case when considering the Pacman model in the parametric case

The dust front ends in particular should be separate anyway, so it's just the stellar / nebular / intrinsic front ends

I think if the docs make it clear that all functionality can be encompassed in a single function, but that there are helper functions, this is okay

WillJRoper (approved):

PR #168, opened 2023-05-30
Fixing the application of a filter transmission curve in filter.apply…

WillJRoper:
…_filter

Fixed the apply_filter method and simplified its use to general axis integration for both rest and observer frame bands. 

## Issue Type

- [x] Bug
- [N.A] Document
- [N.A] Enhancement

## Checklist


christopherlovell (approved):
can't test right now, but seems okay

Issue #155, opened 2023-05-15
Should have the option to add IGM outside flux method on SED


WillJRoper:
Need to add an `apply_igm` method for rare instances where the user wants rest frame luminosities including IGM contribution.

WillJRoper:
This is not possible in the context of the IGM models.

PR #205, opened 2023-06-15
Included method to apply IGM to rest frame SED.

WillJRoper:
Addresses #155. There was no way to apply an IGM transmission curve to the rest frame SED. Now this may be a rare desire but better to include functionality to allow it. This has introduced a single method: `SED.apply_igm_to_lnu`. 

Note, I have adopted the Google docstring format for it. 

## Issue Type

- [x] Enhancement

## Checklist

- [N.A] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/synthesizer/sed.py
+++ b/synthesizer/sed.py
@@ -235,6 +235,19 @@ class Sed:
         if igm:
             self._fnu *= igm.T(z, self.obslam)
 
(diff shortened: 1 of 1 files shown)


WillJRoper:
I'm an idiot 🤦 😂

WillJRoper:
This is actually incompatible, I'm closing the issue and PR.

PR #209, opened 2023-06-16
Update Galaxy classes and spectra methods

christopherlovell:
Major rehaul of the galaxy class structure. 

## The `Galaxy` parent and child classes
The galaxy folder has been completely removed, and replaced with `galaxy.py` which contains the `BaseGalaxy` class. This contains all methods that can be applied to both parametric and particle galaxy types. These specific child classes are contained within `particle/galaxy.py` and `parametric/galaxy.py`. This improves the syntax for importing, as suggested by Will

    from synthesizer.particle.galaxy import Galaxy
    from synthesizer.parametric.galaxy import Galaxy

A warning has been added to `BaseGalaxy` if instantiated on its own to alert users that this should not be used independently, and demonstrating how to import one of the two child classes correctly as above.

## Renamed methods

A number of methods for getting spectra with format `get_*` have been updated to make querying simpler. For example:

`get_intrinsic_spectra` -> `get_spectra_intrinsic`

## Docstrings
(most) docstrings have been changed to Google format, and I've added an interpreter to the sphinx config

## Documentation
Documentation is *currently* being updated to reflect all changes. This PR should fail until all checks on the notebook docs pass.

## Issue Type

- [ x ] Document
- [ x ] Enhancement

## Checklist

- [] I have added docstrings to all methods
- [] My changes generate no pep8 errors


WillJRoper (approved):
All looks good to me. Still a bit of documenting work to do throughout but we can chip away at that. I've added a placeholder method for `generate_lnu` in `BaseGalaxy` which will error if not correctly overloaded. It'll help in the future with development and is a touch more pythonic.

PR #210, opened 2023-06-19
Abundances improvements

stephenmwilkins:
I have improved the clarity of `abundances.py` by calculating the total, gas, and dust abundance patterns. For no depletion gas == total and dust = 0.0. Cloudy should take gas by default since we specify grains later for self-consistency.

Example is updated as is `cloudy.py`.

No documentation yet, but this can be generated from the example. 

## Issue Type

- [ ] Bug
- [ ] Document
- [x] Enhancement

## Checklist

### DIFF
--- a/synthesizer/abundances.py
+++ b/synthesizer/abundances.py
@@ -181,94 +181,158 @@ class Elements:
 
 class Abundances(Elements):
 
-    def __init__(self, Z=Elements.Z_sol, alpha=0.0, CO=0.0, d2m=False, scaling='Dopita+2013'):
+    def __init__(self, Z=Elements.Z_sol, alpha=0.0, C='Dopita2006', N='Dopita2006', d2m=False):
+        
         """
-        This function returns the elemental abundances after removing the depleted amount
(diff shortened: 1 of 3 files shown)


synthesizer/abundances.py
```diff
@@ -181,102 +181,172 @@ class Elements:
...
         self.d2m = d2m
-        self.scaling = scaling
+        
+        # set depletions to be zero
+        self.depletion = {element:0.0 for element in self.all_elements}
 
-        # set all depletions to zero initially
-        self.depletion = {element: 0.0 for element in self.all_elements}
+        # Set helium abundance following Bressan et al. (2012)
+        self.Y = 0.2485 + 1.7756*self.Z
```
aswinpvijayan, line 217: Does the new Helium calibration from Bressan+2012 assume same solar metallicity as ours?

stephenmwilkins: I wondered about this and wasn't sure it should matter. However, it's a very small effect.

aswinpvijayan (approved):
Minor inconsistency in the adopted solar metallicity of Bressan+12 and ours, however the impacts will be minute

PR #214, opened 2023-06-19
Convert vacuum wavelengths to air wavelengths

stephenmwilkins:
Also added a hardcoded list of interesting lines.

## Issue Type

- [ ] Bug
- [ ] Document
- [x] Enhancement

## Checklist


WillJRoper:
Does this not feature in any of the docs? If it does it needs modifying there too.

stephenmwilkins:
> Does this not feature in any of the docs? If it does it needs modifying there too.

I do not believe it does, it's only really used behind the scenes (e.g. when defining line ratios).

However, it is example breaking at the moment because the existing grids use vacuum wavelengths. To fix that I need to recompute the grids but before I do that I want to make other changes to the grid construction.

WillJRoper:
I'd argue to hold this back till then and build your grid modifications off this branch. I'm not approving a merge that actively breaks things. What do you think @christopherlovell, @aswinpvijayan

WillJRoper:
One thumbs-up is enough for me. I'm closing this PR to await the rest of the modifications.

PR #237, opened 2023-06-28
Enhanced the Units implementation and documented their use.

WillJRoper:
This PR introduces the ability for the user to modify the unit system used in a single namespace with Synthesizer. 

It:
- Enables the modification of the default unit system.
- Includes ample warnings to avoid user confusion when modifying the unit system.
- Implements automatic conversion of `unyt_array`s to the unit system upon setting an attribute if the user provides properties with unyt units. 
- Fully documents units.py.
- Included documentation detailing working with units in Synthesizer.
- Moved Filters up in the docs contents for a more logical structure. This puts it above other docs which use filters.

Note: this PR will break some existing examples without #233. This one must be merged first and then the conflicts resolved here. 

## Issue Type

- [x] Document
- [x] Enhancement

## Checklist

### DIFF
--- a/synthesizer/particle/stars.py
+++ b/synthesizer/particle/stars.py
@@ -76,8 +76,7 @@ class Stars(Particles):
     __slots__ = ["metallicities", "nparticles",
                  "tauV", "alpha", "imf_hmass_slope", "log10ages",
                  "log10metallicities", "resampled", 
-                 "velocities", "current_masses",
-                 "s_oxygen", "s_hydrogen"]
+                 "velocities", "s_oxygen", "s_hydrogen"]
 
     # Define class level Quantity attributes
     initial_masses = Quantity()


WillJRoper:
So it turns out certain `numpy` operations break `unyt_array`s. One example being `np.vstack`. These return `unyt_array`s but fail to maintain the unit portion so you are left with a dimensionless `unyt_array`. I'm going to put some logic in to handle these so they are treated as before (as if an `ndarray` had been passed) but this is definitely worth keeping in mind. Either use `numpy` functions with `_` versions of attributes and reassociate units or find other ways that don't break `unyt`.

WillJRoper:
Having dug a bit deeper on the `unyt` repo it would seem they have encountered this before with vstack and have implemented a fix. This implies that there's some issue our end. I wonder if this is what happens when you try and combine incompatible arrays together, it just gracefully combines then without failing but quietly trips the units. 

This would be the case is vstack is fed at least one broken array (spectra in the case I found).  Food for future thought. 

Nonetheless, the dimensionless catch has fixed the issue on this branch.

christopherlovell (approved):

PR #244, opened 2023-07-18
Incorporated index measuring routine.

ConnorSF:
Added the method `calculate_ew` acting on an Sed object in sed.py. 

`def get_equivalent_width` in galaxy.py  calls the method `calculate_ew` after setting the wavelength and flux arrays

 `def calculate_ew` takes two continuum windows and the absorption indices from the spectrum of interest.

An example file `plot_equivalent_width.py` has been added to demonstrate the measuring of index absorption features

An example file `UV_indices.csv` has been added to _../synthesizer/data_ illustrate continuum windows.

## Issue Type

-  Document
- Enhancement

## Checklist


WillJRoper:
Looks great. A couple things:

- Can you implement the same method on `particles.galaxy.Galaxy`?
- We're slowly moving over to Google docstring format (see https://github.com/flaresimulations/synthesizer/blob/main/synthesizer/filters.py) can you finish the doc string here: https://github.com/flaresimulations/synthesizer/pull/244/files#diff-0f8573c254067f24a513e09c80a354d98cb1e87f2b551670bdac3ccc15dcb2d7R503 and convert the other over to google format.

PR #248, opened 2023-07-24
Incorporated index measuring routine with further adjustments.

ConnorSF:
Added the method calculate_ew acting on a sed object in sed.py.

def get_equivalent_width in galaxy.py calls the method calculate_ew after setting the wavelength and flux arrays

def calculate_ew takes two continuum windows and the absorption indices from the spectrum of interest.

An example file plot_equivalent_width.py has been added to demonstrate the measuring of index absorption features.

## Issue Type

- Document
- Enhancement

## Checklist


synthesizer/particle/galaxy.py
```diff
@@ -319,8 +319,6 @@ def get_particle_spectra_stellar(
             the mask.
         """
 
-        from ..extensions.csed import compute_particle_seds
```
christopherlovell, line 322: I think there is a reason these imports are not at the top level, @WillJRoper could you provide?

christopherlovell:
what is the purpose of the `.idea` folder and the `.gitignore` inside it?

ConnorSF:
> what is the purpose of the `.idea` folder and the `.gitignore` inside it?

The .idea folder is automatically generated by the IDE and the gitignore file specifies which files inside the .idea folder should be ignored by Git.

synthesizer/particle/galaxy.py
ConnorSF: If there is a particular reason for not importing at top level, it would be interesting to know why as I have never encountered inline imports before.

christopherlovell:
> > what is the purpose of the `.idea` folder and the `.gitignore` inside it?
> 
> The .idea folder is automatically generated by the IDE and the gitignore file specifies which files inside the .idea folder should be ignored by Git.

Ah okay. It's weird that it doesn't self censor. Could you remove this from the tree?

ConnorSF:
> > > what is the purpose of the `.idea` folder and the `.gitignore` inside it?


Of course. Consider it done

synthesizer/particle/galaxy.py
WillJRoper: With C extensions you often import them in the method that will use that extension. Mostly a separation of concerns approach but also enables you to write alternatives for when the C extensions don't exist. If they were imported at the top level, the whole script would error every time. 

It's common practice but admittedly not PEP-compliant (as far as I am aware). There's method in the madness though.

christopherlovell: @ConnorSF could you put these imports back where they were, then I'll merge this in :)

ConnorSF: @christopherlovell done 👍

christopherlovell (approved):

PR #256, opened 2023-08-16
Add some basic Particle blackhole functionality

### LINKED ISSUE #254: Allow galaxies to have AGN
It would be great if `Galaxy` objects could have a (or multiple) AGN attached to them. When creating images those AGN shouldn't be smoothed other than by the PSF.

### LINKED ISSUE #267: Add method on ParticleGalaxy to calculate the metallicity of the BlackHoles
There needs to be a method on `ParticleGalaxy` that calculates the metallicity of `BlackHoles`, probably by just averaging nearby gas particles or taking the closest gas particle (or both).

stephenmwilkins:
## Issue Type

- [ ] Bug
- [x] Document
- [x] Enhancement

## Checklist

Adds `BlackHoles` object and functionality on Galaxy to store (and use) this object and calculate the black hole metallicity based on surrounding gas particles. Has a skeleton of docs which demonstrates the use of what has been implemented so far.

Other less important additions:

- Introduces the Feltre16 emission model in `generate_grids` from blackhole grid creation.
- Adds `*args` and `**kwargs` to `BaseGalaxy` methods to stop warnings about incorrect arguments due to the overloading.
- Adds black hole units to `Units`.
- Cleans up `get_spectra_stellar`s and `spectra["stellar"]`, in favour of incident naming scheme (sorry! I had unexplainable errors and went too far but its done now).

Closes #267, Closes #254


WillJRoper:
Can you give some explanation of what you've done? This looks nothing like the Stars or Gas implementation. 

Where is the BlackHoles object? The functions in blackholes.py should be methods on this object. Overall I'm confused.

stephenmwilkins:
The two simple functions will ultimately be called by the Parametric and Particle Blackhole classes but to avoid repetition I've put them here.

The `Feltre16` class is currently overkill but basically describes a model incident AGN spectra. There is no analogue of this for stars since stars can't be described by simple forms like this.

WillJRoper:
Have a look at how Stars and Gas are implemented. This is how we should implement the particle side of things. For `ParametricBlackHoles` cases this will necessarily be different but lets implement them separately following what we have already done and then we unify the particle and parametric approaches where we have duplicate functionality. 

If we have black holes implemented differently that will just lead to confusion on both ours and the users part. I'm tempted to close this PR, what do you think @christopherlovell, @stephenmwilkins?

christopherlovell:
> Have a look at how Stars and Gas are implemented. This is how we should implement the particle side of things. For `ParametricBlackHoles` cases this will necessarily be different but lets implement them separately following what we have already done and then we unify the particle and parametric approaches where we have duplicate functionality.
> 
> If we have black holes implemented differently that will just lead to confusion on both ours and the users part. I'm tempted to close this PR, what do you think @christopherlovell, @stephenmwilkins?

Close, or update to use same structure, substituting in the methods

stephenmwilkins:
Yes, I see the need for the class based approach but wouldn't it be better to a base class where methods applicable to both are stored.

- The Feltre16 class is a fundamentally different thing and is used for creating the cloudy grids.
- The method to calculate the BH temperature has entirely disappeared.

WillJRoper:
In the long run probably, but the parametric/particle split introduces a mire of complication in all the galaxy components. For now, it's best to follow what currently exists and implement them separately. Once implemented and functional we can revisit and marry everything if it's clearly the best way. What I want to avoid is an inheritance extravaganza, it increases complexity too much.

> The Feltre16 class is a fundamentally different thing and is used for creating the cloudy grids.
That is why it is commented out, not removed. If needed, it should go elsewhere, probably Felte16.py or preferably a more informative module name. It is obviously different in focus to the front-facing black hole functionality that belongs in that module. 

> The method to calculate the BH temperature has entirely disappeared.
It has not, look at the init. That method took no arguments other than black hole properties now stored on the object. It also only did a single calculation which will never be time-consuming. It is thus automatically calculated at instantiation. No point in having a method for it.

christopherlovell:
Shouldn't the Feltre+16 method go inside a script within the `generate_grids` top level directory?

WillJRoper:
>Shouldn't the Feltre+16 method go inside a script within the generate_grids top level directory?

That would definitely make the most sense.

stephenmwilkins:
OK we can implement them separately and re-merge if needed. I'll start a separate branch for `ParametricBlackHoles`. 

The reason for Feltre+16 being in the top-level was in case it was used elsewhere; though it's probably true that it won't be as we want people to exclusively use the Grids. I'll move it over there for now.

I see the BH T method now. My only issue with calculating it straight away is that some models won't want to use it at all.

Feel free to just adopt this as `ParticleBlackHoles` branch and merge.

synthesizer/grid.py
```diff
@@ -245,7 +245,7 @@ def get_spectra(self):
 
                 # if incident is one of the spectra also add a spectra called "stellar"
                 if spec_name == 'incident':
-                    self.spectra['stellar'] = self.spectra[spec_name]
+                    self.spectra["stellar"] = self.spectra[spec_name]
```
christopherlovell, line 248: if this is going to be turned into a more general grid class, we should be careful to update this

synthesizer/particle/blackholes.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Bolometric luminosity
+        self.bol_luminosity = None
+
+        # Calculate the big bump temperature
+        self.bb_temperature = 2.24E9 * \
+            self._accretion_rate ** (1 / 4) * self._masses ** -0.5
+
+        # The metallicity of the region surrounding the black hole.
+        self.metallicities = metallicities
```
christopherlovell, line 113: I don't think metallicity should be a required argument

christopherlovell (approved):

PR #260, opened 2023-08-21
Dust emission

stephenmwilkins:
## Issue Type

- [ ] Bug
- [ ] Document
- [x] Enhancement

## Checklist

Added dust emission model as well as methods to generate dust emission spectrum using energy balance approach. Examples of both the dust emission model and a parametric model now implemented in docs. 

NOTE: because the test grid is truncated at 20um this example doesn't work particularly well at the moment. I will have to extend the test grid but there is no point doing that until the grid_update PR is accepted.


stephenmwilkins:
I have merged `grid_update` branch and also generated a new test_grid with log10 wavelength spacing extending into the FIR.

stephenmwilkins:
This shouldn't be properly reviewed until the `grid_test` PR is approved since it has been merged with that.

christopherlovell:
Needs updating with main, looks like there are conflicts

docs/source/misc/dust_emission.ipynb
christopherlovell, line 1: Why is this in a separate 'misc' folder? make a new dedicated dust emission section

christopherlovell (commented):
Few minor edits required. Have not tested functionality explicitly

synthesizer/dust.py
```diff
@@ -313,3 +313,229 @@ def attenuate(self, tau_V, lam):
...
+
+
+
+from scipy import integrate
+from unyt import h, c, kb, um, erg, s, Hz
+from unyt import accepts, returns
+from unyt.dimensions import length, time, temperature
+
+
+def planck(nu, T):
```
christopherlovell, line 327: Could this be moved to `utils.py`?

WillJRoper: Have moved

WillJRoper:
I've done some linting in `dust.py` but left the rest for @stephenmwilkins. Lots of cleaning is needed, whitespace around operators, removal of blank lines, and `power_law` needs to be `PowerLaw`, to name a few issues. 

install flake8 and run `flake8 <path/to/file>` on the terminal to see a list of issues.

I'm also somewhat confused by the classes in `dust.py`. None of them have an init so none have attributes. They are essentially a container of functions. This should probably be in files of functions if that is all that is needed.

christopherlovell:
> I'm also somewhat confused by the classes in `dust.py`. None of them have an init so none have attributes. They are essentially a container of functions. This should probably be in files of functions if that is all that is needed.

All the emission derived classes have `__init__`'s with attributes? Or are you talking about the dust curves?

As an aside, I'd argue the emission classes should be moved to their own dedicated script `dust_emission.py`

WillJRoper:
My bad. I somehow missed that the children do. It's the parent that doesn't but it should and the children should instantiate it. 

Agreed on the split of the file.

stephenmwilkins:
> My bad. I somehow missed that the children do. It's the parent that doesn't but it should and the children should instantiate it.
> 
> Agreed on the split of the file.

Yes, we either need `dust_attenuation.py` and `dust_emission.py` or `dust/emission.py` and `dust/attenuation.py`. Not sure what's best?

WillJRoper:
The latter gets my vote. Submodules give a nice structure.

PR #262, opened 2023-08-22
Fix intrinsic spectra

### LINKED ISSUE #261: get_spectra_intrinsic is incorrect
I'm not sure when this happened but the `get_spectra_intrinsic` method in `base_galaxy` is incorrect. It should **not** be the combination of the stellar and nebular spectra. Instead it should be the combination of **transmitted** and nebular. 

Easiest fix is to add a new method `get_spectra_transmitted` which is called by `get_spectra_intrinsic`.

stephenmwilkins:
## Issue Type

- [x] Bug
- [ ] Document
- [ ] Enhancement

## Checklist

Closes #261 

`get_spectra_intrinsic` was combining stellar and nebular spectra which is incorrect. It should be combining transmitted and nebular spectra. This has been fixed by adding a new method to generate the transmitted spectra and modifying `get_spectra_intrinsic` to call this method. I also added a method to generate incident spectra since this is more generic and should supersede `get_spectra_stellar`.


synthesizer/base_galaxy.py
```diff
@@ -64,6 +64,80 @@ def get_spectra_stellar(
...
+    
+    def get_spectra_transmitted(
+        self,
+        grid,
+        young=False,
+        old=False,
+        update=True
+    ):
+        """
+        Generate the transmitted spectra using the provided Grid
```
WillJRoper, line 113: This needs to be more detailed. What is transmitted?

WillJRoper (changes requested):
In principle fine but need to put detail into what exactly "transmitted" is in the doc string. I also don't think the incident renaming is necessary but willing to be convinced otherwise.

synthesizer/base_galaxy.py
christopherlovell: I believe this is the language used by CLOUDY in both its documentation and outputs, and we decided at some point to honour this naming for the sake of simplicity. But I agree that this should be better explained in the docstrings, and ideally have its own section in the documentation

stephenmwilkins:
I have fixed another issue with intrinsic in that it didn't properly account for the escape fraction. It is now the sum of `(1-fesc)*(nebular + transmitted) + fesc * (incident) `. The method also saves the `escape` spectra (i.e. `fesc * (incident)`).

stephenmwilkins:
I have also added a file to the docs describing the different spectra, `spectra.rst`. Ultimately this should be linked to from the various notebooks but since these might change I haven't done this yet.

stephenmwilkins:
I have not fixed the naming issues in the other methods yet.

synthesizer/base_galaxy.py
stephenmwilkins: New doc added to documents.

WillJRoper: Great that we have the docs but this also needs to be in the doc string. Just a brief description, "transmitted, i.e. the incident emission reprocessed by the nebula, ...".

WillJRoper (approved):

PR #277, opened 2023-08-30
fix examples for documentation plotting

christopherlovell:
Fixes some issues with the configuration that enable examples to be displayed in the API.

Also renames the examples to be preceded with `plot_` if they result in a plot, which gets picked up by sphinx-gallery and uses the resulting image in the thumbnail.

## Issue Type
- Document
- Enhancement

## Checklist


christopherlovell:
Also fixes documentation which is currently failing as there is no docstring on the quiescent colours example...

WillJRoper (approved):

PR #280, opened 2023-08-31
Remove dust attenuation

stephenmwilkins:
This simply comments out references to the `dust_attenuation` package. This should allow the entire package to be installable via `pip`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ PyYAML
 scipy
 unyt
 cmasher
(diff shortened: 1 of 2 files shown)


WillJRoper:
How come? Don't we need this package at various points? I assume that is why this PR is failing. I don't see the point in shooting ourselves in the foot to remove the requirements just for a pip installable module we don't want to use yet.

stephenmwilkins:
Implemented in dust_emission PR

PR #305, opened 2023-09-11
Improved Sed

stephenmwilkins:
Re-written much of the `Sed` class and added an example notebook. 

I have also begun consolidating the docs and moving longer examples to examples instead.

## Issue Type

- Bug
- Document
- Enhancement

## Checklist


stephenmwilkins:
I've continued to rationalise the docs and examples removing repetition and consolidating material.

stephenmwilkins:
Hoorah... all checks passed.

synthesizer/sed.py
```diff
@@ -43,17 +36,31 @@ class Sed:
     fnu = Quantity()  # nJy
```
WillJRoper, line 36: Missing `Quantity` for lam and lnu. Also comment is misleading. Units can be changed by the user. The global unit system is respected but doesn't mean it will always be nJy.

synthesizer/sed.py
```diff
@@ -64,28 +71,39 @@ def __init__(self, lam, lnu=None, description=False):
...
             elif self._lnu.ndim == 1:
                 # if single Seds simply add together and return.
                 return Sed(self._lam, lnu=self._lnu + second_sed._lnu)
 
             elif self._lnu.ndim == 2:
-                # if array of Seds concatenate them. This is only relevant for particles.
-                return Sed(self._lam, np.concatenate((self._lnu, second_sed._lnu)))
+                # if array of Seds concatenate them.
+                # This is only relevant for particles.
+                return Sed(self._lam, np.concatenate((self._lnu,
```
WillJRoper, line 96: See issue #317

synthesizer/sed.py
WillJRoper, line 36: Some quantities are missing: z variants and broadband quantities

synthesizer/sed.py
```diff
@@ -64,28 +71,39 @@ def __init__(self, lam, lnu=None, description=False):
...
     def __add__(self, second_sed):
+
+        """
+        Overide addition operator to allow two Sed objects to be added
+        together.
+        """
+
         if not np.array_equal(self._lam, second_sed._lam):
-            exceptions.InconsistentAddition("Wavelength grids must be identical")
+            exceptions.InconsistentAddition("Wavelength grids \
```
WillJRoper, line 81: None of these exceptions actually do anything since they are missing the `raise` keyword. i.e. `raise exceptions.InconsistentAddition(...)`.

WillJRoper: I have now fixed this.

synthesizer/sed.py
stephenmwilkins: I've updated this and changed `nuz` -> `obsnu`. 

Which broadband quantities do you mean? I think things like bolometric_luminosity don't need to be specified as a quantity since we never use the unit-less versions, i.e. they are always returned with units of e.g. `luminosity.units`.

synthesizer/sed.py
stephenmwilkins: Is this fixed now?

synthesizer/sed.py
WillJRoper: I had been poking at the broadband quantities. It would mean modifying a `Quantity` to understand when things are dictionaries. Maybe worthwhile in the long run but I won't bother for now.

synthesizer/sed.py
WillJRoper: Indeed but probably causing issues down stream. I'm in clean up mode from this change.

stephenmwilkins: OK, probably needs a more thorough test suite.

synthesizer/sed.py
stephenmwilkins: OK.

WillJRoper (approved):

PR #337, opened 2023-10-17
Rebin grid spectra

### LINKED ISSUE #335: High resolution wavelength grids are awfully memory hungry
When making an image of a galaxy with 33 million particles I encountered seg faults all over the place. At first, I thought this was the code but it wasn't a single point failing. 

Turns out I ran out of memory either during spectra creation or IFU making (changed depending on the number of elements). This isn't an issue for integrated spectra, however, the moment you have a high-resolution wavelength grid (~88,000 elements for the BPASS grid I used) and lots of particles this will fall apart. 

Proposed solutions:
- Allow the user to apply wavelength cuts to the grid (and thus anything derived from it) to remove elements not used in their current analysis. We carry around a lot of array elements which are effectively 0 compared to the elements with meaningful flux.
- Optional downsampling of spectra for image creation (will smear out lines, so only works when working with photometry and even then not ideal because lines may shift flux outside of the filter curve). 
- Completely rethink the entirety of particle spectra + image creation to improve the memory usage (I obviously don't want to do this. Everything will have to be done on the fly and subsequently deleted when stored in an image. The IFUs also necessitate the current structure).

WillJRoper:
This PR introduces the ability to interpolate the SPS grid spectra onto a user-provided wavelength array at instantiation. This gives the user control over the memory footprint of spectra while enabling consistent use of wavelength arrays specific to certain surveys, instruments, etc.

- Adds `new_lam` argument to `Grid.__init__` (by default None) along with an interpolation mode argument.
- Adds interpolation method to `Grid`.
- Adds missing doc strings and missing attribute to grid.py.
- Closes #335.
- Adds grid interpolation to docs.
- Adds a plotting example of interpolation showing different resolution jumps and interpolation methods.

## Issue Type

- Enhancement

## Checklist


christopherlovell (approved):

aswinpvijayan:
Since the grid has emission lines, when you degrade the spectra, don't you need to use the spectral resampling library here?

WillJRoper:
This has already been merged... but I also don't see what that renin library can be doing better. If you look at the example it's doing a fine job at reproducing the spectra. Maybe more complex for a real grid but I'm not sure the renin library can efficiently rebin along the wavelength axis of an arbitrary dimensional array?

christopherlovell:
Didn't even clock this, apologies. 

The issue is that you need to ensure that the integrated flux is preserved (otherwise things like photmetric fluxes can become biased). This is what [spectres](https://github.com/ACCarnall/SpectRes) does. Not sure if you're referring to a specific 'rebin' library here. 

I suggest raising an issue to fix this. Should be relatively simple to substitute in.

PR #341, opened 2023-10-17
Removing test grid with Google drive Action

### LINKED ISSUE #272: test grid smears out lines
Test grid creation smears out emission lines to several bins which is undesirable. This may need some thinking about how best to do the downsampling.

WillJRoper:
This PR deletes the test grid and replaces it with a GitHub action that installs `gdown` (a pip installable interface to Google Drive) and downloads a bpass grid from [here](https://drive.google.com/file/d/1txY9PY-ejtRaMSZorpZSxGPsm6muvh9X/view?usp=sharing). 

This test grid is named `test_grid.hdf5` and stored in `tests/test_grid` meaning all examples still run without issue (modulo a couple of imaging ones that needed modifying to account for the extremely high wavelength resolution of the BPASS grid).  We can change the grid downloaded to any grid we want and the action will always rename it test_grid.hdf5.

This required adding a Google Drive API key to the repo's secrets. We can reuse this API key to store any data we want and download it in the action.

A user can also download the test grids via the `tests/get_test_grid.sh` script. This script will install `gdown` first (harmless if already installed) and will then download the test grid storing it at the expected location.

Closes #272 by using a real grid for the tests.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -32,7 +32,7 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
         pip install -e .
         # Install google API for downloading test data
-        pip install --upgrade google-auth google-auth-oauthlib google-auth-httplib2 google-api-python-client
+        pip install --upgrade google-auth google-auth-oauthlib google-auth-httplib2 google-api-python-client gdown
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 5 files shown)


christopherlovell:
Fantastic. Could we also look into removing the test grid from the git history, to reduce the bloat?

WillJRoper:
Absolutely! This will need merging first and then I'll under take the scary process 😂

Any other data you'd like treated like this?

christopherlovell:
There are some methods for checking the biggest files in the git history, there may be some old grids Steve created lurking about...

Will approve and merge this.

WillJRoper:
Those have already been killed off, I think anyway. We could include some real galaxy data this way for the camels examples and the imaging?

christopherlovell:
Let's merge this in for now. I don't think we need any cosmo data for testing purposes. It could be useful when running the docs, however we don't want to include arbitrarily large data in the github workflow as it will slow down the testing and doc compilation.

christopherlovell (approved):

PR #345, opened 2023-10-18
Introduced a Quantity for Grid.lam

WillJRoper:
Does what it says on the tin. Nothing more, nothing less

## Issue Type

- Enhancement

## Checklist


WillJRoper:
Note that this requires #344 to resolve the final issue since they are related. Using the `grid.lam` in the `Filter` instantiation fails because of the units which are correctly handled by the changes in #344. No point duplicating the changes here.

christopherlovell (approved):

PR #349, opened 2023-10-24
SC-SAM example

louiseseeyave:
## Issue Type
- Document


christopherlovell:
I think we may want to remove the example file from the repo due to its size, and add it to the optional downloads from gdrive

christopherlovell (approved):

PR #358, opened 2023-10-30
Grid fixes

christopherlovell:
Main change is to add a `check_lines_available` method to `grid.py`, which is run before loading lines to check the grid file contains them. Raises an informative exception if not. We may wish to change the default read_lines flag to False.

- small pep8 fixes (docstring updates, removed unused methods)
- added exception to `flatten_linelist` method
 

## Issue Type
- Bug

## Checklist

### DIFF
--- a/synthesizer/grid.py
+++ b/synthesizer/grid.py
@@ -20,15 +20,17 @@ def get_available_lines(
     """Get a list of the lines available to a grid
 
     Arguments:
-        grid_name (str)
+        grid_name (str):
             The name of the grid file.
 
-        grid_dir (str)
+        grid_dir (str):
             The directory to the grid file.
 
     Returns:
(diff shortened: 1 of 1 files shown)


stephenmwilkins:
Any idea of the remaining issue here? I realise this bug has broken some of my code.

christopherlovell:
@stephenmwilkins the remaining failing test is fixed in #362

WillJRoper:
@christopherlovell: Want to set read_lines=False in this PR? That makes sense to me. More intentional that way!

christopherlovell:
I'm not sure, for new users it might be more confusing when they go to access line information and it's not there. At least this way it makes it explicit that you don't have lines

WillJRoper (approved):

PR #362, opened 2023-10-30
Reducing memory overhead in plot_composite_galaxy

### LINKED ISSUE #359: Bug in composite images example leading to failing docs
`examples/parametric/plot_composite_ga;axy.py` currently fails on line 106 with a segfault:

    
    # make images
    bulge_img = bulge.make_images(
        resolution=resolution,
        filters=filters,
        sed=bulge.spectra["incident"],
        fov=fov,
    )

To reproduce, run this example. Should run without issues. The disk imaging example just above this line runs without issue, so unclear what is unique about this particular instance.

WillJRoper:
Could Close #359 if the citied segfault is indeed to do with the memory available as I'm guessing. Loading the new test grid the images would be ~(50, 50, 80000). Overkill for this example. If this PR does not error then this is almost certainly the case and this PR does close the issue.

The fix is to pass a lower resolution wavelength grid at `Grid` instantiation. As in other imaging examples. Sadly this will always be a necessary evil for IFU creation. Although it is computationally more efficient to make the IFU and make images from it we may want to pivot and only make IFUs specifically when they are wanted. 

## Issue Type

- Bug

## Checklist

### DIFF
--- a/docs/source/parametric/generate_composite_galaxy.ipynb
+++ b/docs/source/parametric/generate_composite_galaxy.ipynb
@@ -67,7 +67,7 @@
    "source": [
     "grid_dir = '../../../tests/test_grid'\n",
     "grid_name = 'test_grid'\n",
-    "grid = Grid(grid_name, grid_dir=grid_dir)"
+    "grid = Grid(grid_name, grid_dir=grid_dir, new_lam=np.logspace(3, 4.2, 1000))"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


stephenmwilkins:
Just so I understand, is the issue that we have too many wavelength points resulting in a huge 3D grid? 

I'd resist using a lower resolution grid just because then we lost the ability to explore dynamics. A 1000 km/s offset is only 17\AA.

As an alternative we might like to add a method to `Grid` that restricts the wavelength range of the grid somehow.

christopherlovell:
> As an alternative we might like to add a method to Grid that restricts the wavelength range of the grid somehow.

@stephenmwilkins this is exactly what already exists and what Will has done to fix this issue in this particular example. The issue is not in the test grid, but in the way that images are created via an IFU over all wavelength points.

@WillJRoper I agree that we should modify image creation to take either a single wavelength or some restricted set, rather than all via an IFU by default.

christopherlovell (approved):

PR #366, opened 2023-10-31
Update cloudy.py

stephenmwilkins:
Turned off print statement in read_linelist function.

## Issue Type

- Bug
- Document
- Enhancement

## Checklist

### DIFF
--- a/synthesizer/cloudy.py
+++ b/synthesizer/cloudy.py
@@ -498,7 +498,7 @@ def read_linelist(filename, extension="elin"):
             # convert luminosity to float
             lum = float(lum)
 
-            print(id, wavelength, lum)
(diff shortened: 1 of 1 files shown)


christopherlovell (changes requested):
Could you add a verbose flag argument to the method, turn it off by default, and then print this if enabled?

stephenmwilkins:
> Could you add a verbose flag argument to the method, turn it off by default, and then print this if enabled?

I really don't think it's essential, I think it was only there during some bug searching.

christopherlovell (approved):

christopherlovell:
I'd say delete the line rather than commenting it out then, but will merge now

PR #369, opened 2023-11-01
Photoionisation improvements

stephenmwilkins:
Making photoionisation modelling more general (i.e. for different cloudy versions and possible other photoionisation codes).

- created `photoionisation/`
- moved current `cloudy.py` to this folder and renamed `cloudy17.py`
- moved generic functions to `photoionisation.py`
- some other tidying


synthesizer/photoionisation/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .photoionisation import *
```
WillJRoper, line 1: Why are you importing everything and then importing cloudy17 explicitly below? We should avoid * imports in favour of clarity.

WillJRoper: Also we shouldn't be using relative imports

WillJRoper (changes requested):

stephenmwilkins:
With the change in structure I somehow managed to destroy this branch so closing it down and starting again.

Issue #373, opened 2023-11-02
Add `sed.concat` to the docs


christopherlovell:
`sed.concat()` is a really useful method on the sed class. Should be added to the docs, ideally somewhere [here](https://flaresimulations.github.io/synthesizer/sed.html#Multiple-SEDs)

Should include example with a list of seds, e.g. `sed.concat(*seds)`

PR #374, opened 2023-11-02
Fixed typo in sed.concat

WillJRoper:
np.concatanate(x, y) -> np.concatenate((x, y))

Also fixes unut related errors that have appeared in #372 and this PR. The fix for these was to use the unitless version of attributes in several filter functions and in the blackhole grid docs. I presume a package updated changing the behaviour in these situations. Particularly when using `np.trapz` but also `SED(lam, lnu)` so probably a unyt update.

## Issue Type

- Bug

## Checklist


christopherlovell:
do you want to fix #373 here too?

christopherlovell (approved):

Issue #225, opened 2023-06-21
Add git revision to grid file


christopherlovell:
Record the git revision in the grid file on creation.

see second answer [here](https://stackoverflow.com/questions/14989858/get-the-current-git-hash-in-a-python-script)

WillJRoper:
It just dawned on me that the second answer in the linked SO post doesn't work outside the repo without some "inventive" `os.cd`ing. What we actually need to do is bake this into the module in some way such that `synthesizer.git_hash` (or similar) returns the git hash of the installed version. Then we can attach this to Grids to version them. 

It seems it may be a bit more complex than I first thought but certainly doable. Prior to PyPI upload, we can have this attribute set at install time using the method in the SO above since this will only happen once but probably best to find a future proof solution.

WillJRoper:
Or... we just fully adopt intense versioning with major.minor.n_pull_request where every accepted PR increments the final number and then store that. It's less automatic but less of a headache.

WillJRoper:
Answer 6 on https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package addresses github revisions and automating versioning based on it. Most answers I've found (as this one states) ignore the existence of version control entirely. 

It seems the broad approaches are:
- Have `__version__` defined in `setup.py`. I've seen mention that this leads to `__version__` being unimportable in the package, exactly what we want, and I can't see how you would import it if defined there...
- Have a `_version.py` script that defines the `__version__`, `import __version__ from _version` in the main package `__init__.py`, and `exec(open('synthesizer/_version.py').read())` in `setup.py` to make it accessible there.

The latter seems complex but very functional. There's a lot of disagreement on how this should all be handled with multiple PEPs rejected based on it. Lots of packages follow the rules detailed in rejected PEPs too. Since we actively need this versioning for functionality I'd be keen to implement a version of `__version__` from a rejected PEP. The latter approach makes the most sense for us. 

We could ignore the git integration detailed in the SO thread. It's pretty OTT, the downside would be simply having to ensure the git tags match `__version__` in `_version.py`.

WillJRoper:
Argh... all that said there is also the simple:
```
from importlib.metadata import version
version('wheel')
# '0.33.4'
```
but only for Python >=3.8. I *assume* this uses the version number (already) defined in `setup.py` (https://github.com/flaresimulations/synthesizer/blob/main/setup.py#L115). Again, this would require syncornisation with git releases but... not a huge deal.

[...]

PR #380, opened 2023-11-07
Updating installation with pyproject.toml 

### LINKED ISSUE #224: Add version number to setup.py, create Revision on Github
As it says on the tin.

WillJRoper:
This PR will remove the use of legacy and deprecated installation methods.

- Removes all mentions of DistUtils in favor of setuptools versions.
- Introduces pyproject.toml.
- Cleans up setup.py to only be used for extension compilation (something not yet supported by the pyproject.toml method.
- Automatic git-integrated version numbers from setuptools_scm.
- Removal of numpy egg in favor of `build_requires` in pyproject.toml.
- Closes #224.
- Provides the mechanism to add git information in the version number to grid files to address #225.
- To make editable installs work the package has been converted to "src" structure. This somewhat cleans up the root of the repo with all non-example code now contained in the `src` directory.

These changes will also require some modification of the installation workflow but that has to be done directly on `main`.

This PR is in draft format because:
- [x] The name will be what is used on pip, we need to decide on that and I've forgotten the outcome of previous discussions.
- [x] Are we happy with the automatic versioning?
- [x] Metadata for authors etc needs checking.
- [x] Are we happy with the README being used for the long description? (See `readme` key in pyproject.toml).

In the future, we need to settle on compilation flags, maybe handle the mire that is Windows, and flesh out the test suite of compilation flags in setup.py. 

There is a big commented block I've left related to OpenMP in setup.py. Until the threading is settled I'm leaving this commented, I want to look into pthreads before removing it. A future PR will clean up this commented block and the compilation flags. 

## Issue Type

- Enhancement

## Checklist


pyproject.toml
```diff
@@ -0,0 +1,131 @@
...
+  "synthesis",
+]
+
+# Synthesizer authors
+authors = [
+  {name = "C. S. Fournier", email = "connor.sant.17@um.edu.mt"},
+  {name = "C. C. Lovell", email = "christopher.lovell@port.ac.uk" },
+  {name = "W. J. Roper", email = "w.roper@sussex.ac.uk" },
+  {name = "L. Seeyave", email = "L.Seeyave@sussex.ac.uk"},
+  {name = "A. J. Vijayan", email = "apavi@space.dtu.dk" },
```
aswinpvijayan, line 47: Could you change that 'J' to a 'V' here and below

WillJRoper: 😂 no idea how that slipped through

WillJRoper: Done

aswinpvijayan (approved):

PR #385, opened 2023-11-08
Instead of caching filters we provide write/load methods

### LINKED ISSUE #242: Instead of caching filters we provide write/load
**Is your feature request related to a problem? Please describe.**
If an SVO database connection can't be made we currently can't use "real" filter curves. We also don't want a cache in the background secretly writing to a users system.

**Describe the solution you'd like**
Provide a method to write a `FilterCollection` to a HDF5 File and a function to load a `FilterCollection` from a HDF5 file. This is in the control of the user and provides a "caching" of these online filters.

WillJRoper:
Added the ability to write `FilterCollection`s to HDF5 files and read them back in at instantiation of a new `FilterCollection`. 

Added a section to the docs detailing this.

This closes #242.

One thing to note is that a `FilterCollection` passed a path to load a file will ignore any other arguments passed. The code will warn the user if they try this. This is because it becomes complicated with wavelength arrays otherwise and simpler to warn the user. They can create `FilterCollection`s separately and add them without issue. There are then a host of baked-in errors to ensure they don't do anything silly. I felt this was a safer approach and simpler in the backend. 

The units are saved to the file ensuring consistency in the unit system. Even if the user has redefined the unit system the units will be correctly converted to their current unit system.

The version is saved alongside the `Filter` data and the code will raise a warning if the file and code versions don't match.

## Issue Type

- Enhancement

## Checklist


christopherlovell (changes requested):
Why is the `_load_filters` method hidden (i.e. preceded with an underscore)?

christopherlovell:
gotcha, misunderstood, looks good

stephenmwilkins:
I love HDF5 files but since SVO provides ascii files wouldn't it be good to also be able to point to a directory of SVO ascii files. It would be nice to also have a help function to download multiple SVO ascii files.

stephenmwilkins:
> The way it's structured now is you save a whole `FilterCollection`, obviating the need for a specific helper function. I can see advantages to having that flexibility, but also the elegance of a single hdf5 file. If you need to add more filters at a later date, you can just load this file, add another filter, and save again.

I see that but I think it would still be good to be a load a filter from ascii and/or a folder. I'll raise it as a separate issue though for us to discuss. I have no objections to doing this.

WillJRoper:
Personally, I don't see the bonus of adding multiple saving methods. It just gets confusing providing multiple ways to do stuff.

stephenmwilkins:
> Personally, I don't see the bonus of adding multiple saving methods. It just gets confusing providing multiple ways to do stuff.

It's not saving methods, it's reading methods. There will be some filters (e.g. instruments not yet commissioned) that won't be on SVO so we should be able to read those as well.

WillJRoper:
That's what the generic filter type is for. The user can supply arbitrary transmission and wavelength arrays. 

We can never predict exactly how a user will have those stored. Better to leave the user to load those and pass them as arrays. If we had an accepted input format for ASCII the user would have to convert anyway so they'd end up doing more work than just handing arrays.

christopherlovell (approved):

PR #397, opened 2023-11-10
Move load data scripts to dedicated folder

### LINKED ISSUE #360: Should use dust mass as a gas particle property, not DTM
Currently, we store dust-to-metals ratio on each gas particle, and use this to calculate the dust mass by combining with the gas particle mass and metallicity (e.g. in the LOS calculation). Some simulations, e.g. Simba, have a live dust model that explicitly follows the dust mass. We should, by default, store the dust mass on each particle; for simulations without this property, this mass should be calculated on load in assuming some dust-to-metals ratio. This will simplify all downstream tasks and imaging. 

Thought / comments welcome

christopherlovell:
Moves all load data scripts to their own dedicated folder, to tidy up the root source directory. All new load data scripts should go here. This folder could also contain scripts for loading other kinds of data e.g. powderday outputs.

- created `load_data` folder
- renamed `load_sam_data` to `load_scsam.py`
- flares, simba and camels scripts added
- `utils.py` contains useful utility methods for different simulations
- closes #360 

## Issue Type
- Enhancement

## Checklist


WillJRoper (changes requested):
All looks good. Only a small request:

Can you add dust_masses to the unit system and make it a `Quantity`? Also need to add to the PR that this closes that issue.

WillJRoper (approved):

PR #398, opened 2023-11-13
Calzetti fix

### LINKED ISSUE #142: Calzetti dust won't work
The implementation of the Calzetti dust law uses the N09 class from the external `dust_attenuation` module. However, this throws an exception if the wavelength is outside a particular range (which ours are). We need wrap some code around this that avoids the exception. For example, we could generate the attenuation curve only over the allowed portion (setting the rest to 1.0)?

aswinpvijayan:
## Issue Type
- Enhancement
Changed the Calzetti attenuation curve definition to remove infinities and nan at longer wavelengths.
I have also removed the from the requirements dust_attenuation package which is no longer used.

Closes #142 

## Checklist

### DIFF
--- a/src/synthesizer/dust/attenuation.py
+++ b/src/synthesizer/dust/attenuation.py
@@ -43,28 +43,35 @@ def N09_tau(lam, slope, cent_lam, ampl, gamma):
     lam_v = 0.55
     k_lam = np.zeros_like(lam_micron)
 
-    ok1 = (lam_micron >= 0.12) * (lam_micron < 0.63)
-    ok2 = (lam_micron >= 0.63) * (lam_micron < 31.)
-    ok3 = (lam_micron < 0.12)
-    if np.sum(ok1) > 0:
+    # Masking for different regimes in the Calzetti curve
+    ok1 = (lam_micron >= 0.12) * (lam_micron < 0.63) # 0.12um<=lam<0.63um
+    ok2 = (lam_micron >= 0.63) * (lam_micron < 31.) # 0.63um<=lam<=31um
+    ok3 = (lam_micron < 0.12) # lam<0.12um
+    if np.sum(ok1) > 0: # equation 1
         k_lam[ok1] = -2.156 + (1.509 / lam_micron[ok1]) \
             - (0.198 / lam_micron[ok1]**2) \
             + (0.011 / lam_micron[ok1]**3)
         func = interpolate.interp1d(lam_micron[ok1], k_lam[ok1],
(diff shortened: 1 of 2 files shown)


src/synthesizer/dust/attenuation.py
```diff
@@ -39,26 +39,35 @@ def N09_tau(lam, slope, cent_lam, ampl, gamma):
...
-    k_lam = np.zeros_like(tmp_lam)
-
-    ok = (tmp_lam < 0.63)
-    k_lam[ok] = -2.156 + (1.509 / tmp_lam[ok]) \
-        - (0.198 / tmp_lam[ok] ** 2) \
-        + (0.011 / tmp_lam[ok] ** 3)
-    k_lam[~ok] = -1.857 + (1.040 / tmp_lam[~ok])
+    k_lam = np.zeros_like(lam_micron)
+
+    ok1 = (lam_micron >= 0.12) * (lam_micron < 0.63)
```
christopherlovell, line 46: could you provide some brief comments here as to what the `ok1`, `ok2` and `ok3` masks represent?

src/synthesizer/dust/attenuation.py
```diff
@@ -39,26 +39,35 @@ def N09_tau(lam, slope, cent_lam, ampl, gamma):
...
+            + (0.011 / lam_micron[ok1]**3)
+        func = interpolate.interp1d(lam_micron[ok1], k_lam[ok1],
+                                fill_value="extrapolate")
+    if np.sum(ok2) > 0:
+        k_lam[ok2] = -1.857 + (1.040 / lam_micron[ok2])
+    if np.sum(ok3) > 0:
+        k_lam[ok3] = func(lam_micron[ok3])
+    
     k_lam = 4.05 + 2.659 * k_lam
+    k_v = 4.05 + 2.659 * (-2.156 + (1.509 / lam_v)
```
christopherlovell, line 61: Do these values come from Noll+09 or Calzetti+00 ?

christopherlovell (changes requested):

WillJRoper:
Does this close #142?

src/synthesizer/dust/attenuation.py
aswinpvijayan: Added description

src/synthesizer/dust/attenuation.py
aswinpvijayan: These are from Calzetti, the UV bump is the addition from Noll+09

christopherlovell (approved):

PR #406, opened 2023-11-14
Fix age mask units

christopherlovell:
Currently the age mask on a stars particle object compares in years, but is expecting `young` / `old` to be in Myr. 
- added method to check the units of `young` and `old` ni all the specrta generation methods
- fixes a bug in `_prepare_sed_args` where the particle number was set to the original array size, not the size of the mask
- docstring updated

## Issue Type
- Bug

## Checklist

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -6,6 +6,7 @@ and methods common between them.
 """
 import numpy as np
 import matplotlib.pyplot as plt
+from unyt import Myr, unyt_quantity
 
 from synthesizer import exceptions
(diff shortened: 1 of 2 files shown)


WillJRoper (commented):
Can this not use a `unyt` conversion to be more explicit?

christopherlovell:
Sure, arguably binning the `log10age` attribute would make this easier

WillJRoper:
Agreed, but thats the units of the `Grid` and I believe the SPS model axis?

WillJRoper (approved):

PR #413, opened 2023-11-15
Improvements to spectra plotting methods and galaxy instantiation fix

### LINKED ISSUE #273: plot_spectra method on Galaxy could be improved
This method isn't very nice at the moment. As part of PR #260 I've made some improvements including automatically setting the x-limits based on the spectra to be plotted (e.g. only UV-->NIR for most spectra) and setting one way of setting the y-limits.

WillJRoper:
This PR improves the helper methods for plotting spectra.

- Generalised plotting functions are now in `sed.py` for rest frame spectra (`plot_spectra`) and observed spectra (`plot_observed_spectra`). The latter is a wrapper around the former so modifications to spectra plotting can be made in a single place.
- All helper methods are now a wrapper to these functions.
- Stellar spectra can be plotted for singular `Sed`s, a list of specified keys or all spectra based on the inputs to `Stars.plot_spectra` (and observed equivalent).
- The galaxy helper method is ready to handle multi-component spectra now in the same way as described above and is capable of plotting specified spectra for any combination of components (or indeed all spectra). 
- If limits are not provided to any of these methods/functions the code attempts to find reasonable values without plotting 20 orders of magnitude. In fact, it aims for a lower limit 5 magnitudes below the peak which seems to work well enough. Realistically the user should provide their own limits if a better set is desired. 
- Units for labels can be specified by arguments (applicable for when spectra are per stellar mass and that isn't included in the unit system explicitly) but are by default automated via extraction from the `unyt` units attached to `Sed` attributes and "prettified".
- Labels are now also made prettier automatically or can be specified explicitly in the case of a single spectra object.
- Although not applicable in any of the helper methods the user can pass their own figure and axes objects to plot on. This can be useful when using these methods to plot onto subplots.

In addition to the above, I have fixed the instantiation of galaxies. They previously didn't call the `__init__` of `BaseGalaxy` and all in common attributes we duplicated across Galaxies. To get the observed spectra photometry plotting to work this needed fixing. 

Closes #273 

## Issue Type

- Enhancement

## Checklist
- [x] I have added tests that prove my fix is effective or that my feature works (already implemented)


src/synthesizer/components/stellar.py
```diff
@@ -985,128 +985,124 @@ def check_young_old_units(self, young, old):
...
+        show=False,
+        ylimits=(),
+        xlimits=(),
+        figsize=(3.5, 5),
     ):
         """
-        plots all spectra associated with a galaxy object
+        Plots either a specific spectra or all spectra provided in a dictionary.
+
+        This is a generic plotting function to be used either directly or to be
```
christopherlovell, line 998: This and other docstrings need updating, this is _not_ the generic plotting function

christopherlovell:
The only thing I'm unsure about here is the `plot_observed_spectra` method on the `stars` component. A stars object shouldn't have an associated redshift, this should be taken from the parent galaxy object. As such, any discussion of observed quantities should be with respect to galaxies, and components should not have  `plot_observed_spectra` methods.

christopherlovell (changes requested):

src/synthesizer/components/stellar.py
WillJRoper: 🤦 indeed! Now done!

WillJRoper:
I agee with the component observed spectra comment. I have removed the method!

christopherlovell (approved):

PR #419, opened 2023-11-16
Rename cloudy parameters

### LINKED ISSUE #417: Rename parameters in abundances.py and cloudy27.py to be PEP-8 compliant
Rename parameters in `abundances.py` and `cloudy27.py` to be PEP-8 compliant.

### LINKED ISSUE #418: Add an abundances notebook
Add an abundances notebook to the docs. I thought there was one, but maybe that was just an example?

stephenmwilkins:
This began by renaming parameters in `photoionisation/cloudy17.py` and `abundances.py` to be more description (e.g. `U` -> `ionisation_parameter`) but grew to include a breaking change to `Abundances` and the addition of an abundances notebook, superseding the example. Specifically, instead of providing a Nitrogen or Carbon abundance via arguments (etc. nitrogen_abundance) you now pass an `abundance` dictionary. I also made some formatting tweaks and accuracy improvements.

Closes #417 and closes #418 

## Issue Type

- [x] Enhancement

## Checklist


src/synthesizer/abundances.py
```diff
@@ -245,118 +326,137 @@ class Elements:
 
 class Abundances(Elements):
     def __init__(
-        self, Z=Elements.Z_sol, alpha=0.0, C="Dopita2006", N="Dopita2006", d2m=False
+        self,
+        metallicity=Elements.solar_metallicity,
+        alpha=0.0,
+        abundances=False,
+        carbon_abundance=False,
```
christopherlovell, line 333: Could you add these args to the docstring

src/synthesizer/abundances.py
```diff
@@ -16,7 +16,88 @@
...
+class ScalingFunctions:
+
+    """
+    This is a class holds scaling functions for individual elements. In each
+    case the function returns the logarthimic abundance relative to Hydrogen
+    for a given metallicity.
+    
+    Example:
+
+    ScalingFunctions.N.Dopita2006(0.016)
```
christopherlovell, line 34: A given publication / model may include many scaling relations, so wouldn't it be better to have `ScalingFunction.Dopita2006.N`  ?

src/synthesizer/abundances.py
```diff
@@ -245,118 +326,137 @@ class Elements:
...
+        self.carbon_abundance = carbon_abundance
+        self.nitrogen_abundance = nitrogen_abundance
+        self.dust_to_metal_ratio = dust_to_metal_ratio
 
         # set depletions to be zero
         self.depletion = {element: 0.0 for element in self.all_elements}
 
-        # Set helium abundance following Bressan et al. (2012)
-        self.Y = 0.2485 + 1.7756 * self.Z
+        # Set helium mass fraction following Bressan et al. (2012)
```
christopherlovell, line 364: Could we add a DOI for this paper? and if possible a description of the values (are these just linear fits?)

christopherlovell (changes requested):

src/synthesizer/abundances.py
stephenmwilkins: Possibly, though I think it's less intuitive to call then.

src/synthesizer/abundances.py
stephenmwilkins: I'm confused, they are there aren't they?

src/synthesizer/abundances.py
stephenmwilkins: However, that would allow us to add an additional attribute, like the paper DOI. So, yes, I think you're probably right.

src/synthesizer/abundances.py
stephenmwilkins: Added

src/synthesizer/abundances.py
stephenmwilkins: I've done this now.

christopherlovell (approved):

PR #424, opened 2023-11-17
Simplify the arguments for parametric.Stars

### LINKED ISSUE #420: Use a single argument on parametric `Stars` to accept sfh/Z function or array
There are currently two separate arguments on `Stars` to accept either functional forms of the SFH/Z distribution (`sf_hist_func` / `metal_dist_func`) and two arguments for arrays (`sf_hist` / `metal_dist`). Can we combine these, and have some logic in the background to interpret what the user provides?

The number of arguments could also be reduced / simplified by removing the `instant_sf` and `instant_metallicity` options, as these arguably can be accomodated within distributions, but no strong feelings on this point.

WillJRoper:
Instead of arguments for each possible type of input to `Stars`, there is now only `sfzh` for providing SFZH arrays, and `sf_hist` and `metal_dist` which can be either instantaneous values, arrays, or function classes for defining the SFH and metallicity distribution respectively. 

Internally the same attributes for each case exist as before to make the parsing and production with each method simpler. This also signposts clearly what is stored in each attribute rather than having an attribute that can be multiple possible data types. 

All examples and docs have been updated.

Closes #420 (Noice!).

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/examples/imaging/image_addition.py
+++ b/examples/imaging/image_addition.py
@@ -36,10 +36,10 @@ stellar_mass = 1e10
 stars = Stars(
     grid.log10age,
     grid.metallicity,
-    instant_sf=10.,
-    instant_metallicity=0.01,
+    metal_dist=10.0,
+    metal_dist=0.01,
     initial_mass=stellar_mass,
-    morphology=morph
+    morphology=morph,
 )
 
 # Get galaxy object
--- a/docs/source/galaxies/galaxy_obj.ipynb
+++ b/docs/source/galaxies/galaxy_obj.ipynb
@@ -46,7 +46,7 @@
     "sfh = SFH.Constant(duration=100 * Myr)\n",
     "\n",
     "# Initialise the parametric Stars object\n",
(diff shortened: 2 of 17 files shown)


examples/imaging/image_addition.py
```diff
@@ -36,10 +36,10 @@
 stars = Stars(
     grid.log10age,
     grid.metallicity,
-    instant_sf=10.,
-    instant_metallicity=0.01,
+    metal_dist=10.0,
```
christopherlovell, line 39: duplication?

christopherlovell (changes requested):

examples/imaging/image_addition.py
WillJRoper: Surprised that wasn't picked up in the linting!

WillJRoper: Fixed now

christopherlovell (approved):

PR #426, opened 2023-11-17
Warning/Error for particles outside grid

### LINKED ISSUE #239: Warn when particles lie outside SPS grid axes
**Is your feature request related to a problem? Please describe.**
The code should produce a warning when particles lie outside the SPS grid, this protects against hidden issues when sampling is forcibly truncated at the grid boundaries.

**Describe the solution you'd like**
A warning, not an error, so the user has feedback but the code doesn't fail on edge cases that don't matter.

WillJRoper:
This PR will introduce a check that particle properties lie within the grid axes (currently not adapted for arbitrary axes because that would require pluralised `Grid` attributes which need fixing on a separate PR addressing the arbitrary axes modifications). 

The code checks how many particles lie outside the grid range. Then:
- If 50% or more of particles lie outside the grid an error is raised.
- If >0% but <50% are outside a warning is printed.

From my tests, this check appears to have minimal impact on performance (it's a couple of Numpy reduces in the background to get the masks). However, I'm tempted to add an argument to flag a skip of this check for users with lots of particles and knowledge their grid and particles are fine. What do people think?

Closes #239 

## Issue Type

- Enhancement

## Checklist


christopherlovell:
I don't think we need to be throwing a hard error when some ultimately arbitrary fraction of particles lie outside the grid range. Users may specify some extreme examples as test cases (e.g. isolating the effect of very young stars) and this will prevent these kinds of experiments. We should just calculate these fractions if asked and report them if the verbose flag is set. This way if things look weird, people can do a quick check of this for production runs.

Also there seem to be some unrelated changes to the sc-sam example that have crept in

src/synthesizer/particle/stars.py
```diff
@@ -363,12 +362,38 @@ def generate_lnu(
...
+            grid.metallicity[0] < self.metallicities,
+        )
+        inside_mask = np.logical_and(age_inside_mask, met_inside_mask)
+        n_out = self.metallicities[~inside_mask].size
+        ratio_out = n_out / self.nparticles
+
+        # Tell the user if there are particles outside the grid, or worst case
+        # error if too many are outside
+        if ratio_out > 0.5:
+            raise exceptions.GridError(
```
christopherlovell, line 388: still raising an exception here?

christopherlovell (changes requested):

src/synthesizer/particle/stars.py
WillJRoper: 🤦 sorry, I only pushed the change to the particle version. Pushed both now

WillJRoper: Pushed now (had to also handle a local conflict because of the merge I did here).

christopherlovell (approved):

PR #428, opened 2023-11-20
Fix attenuation n18

christopherlovell:
Fixes data directory for `npz` files for the Milky Way attenuation in Narayanan+18. also updates example script to plot these.

We should look into extracting the relevant data into plain text files, as npz files are a security risk.

## Issue Type
- Bug

## Checklist
- [x] I have read the [CONTRIBUTING.md]()


WillJRoper (approved):
Agreed, we should look at a better format for these!

PR #430, opened 2023-11-20
Update pacman to accept non-cloudy grids

christopherlovell:
Fixes bugs in the pacman model where non-cloudy grids will not run.

## Issue Type
- Bug

## Checklist


WillJRoper (approved):
I see why this irked you but I think its a necessary evil this way. Better than completely falling over in the pacman model for these grids.

PR #434, opened 2023-11-21
Update simba loading

christopherlovell:
## Issue Type
- Enhancement

## Checklist
- [] I have added docstrings to all methods
- [] I have added sufficient comments to all lines
- [] My changes generate no pep8 errors
- [] New and existing unit tests pass locally with my changes


WillJRoper (commented):
Happy to leave it if it may be necessary but do you need the old star forming mask that you've left commented out with the new better definition?

christopherlovell:
the new definition is not necessarily 'better', just different and arguably more appropriate for Simba given that it gives you the H2 masses explicitly.

Have removed reference to previous, and updated comment to be more explicit.

WillJRoper (approved):

PR #436, opened 2023-11-21
Fixed the particle imaging docs.

WillJRoper:
Fixed the issue in the particle imaging documentation. For some reason the units I had added to automatically convert everything to the global system had been removed (I assume in the move). Thus the kpc units of the loaded data were assumed to be Mpc and the images failed completely.

## Issue Type

- Bug

## Checklist

### DIFF
--- a/src/synthesizer/imaging/scene.py
+++ b/src/synthesizer/imaging/scene.py
@@ -548,10 +548,10 @@ class ParticleScene(Scene):
 
         # Calculate the centre if necessary
         if self.centre is None:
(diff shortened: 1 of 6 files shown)


christopherlovell:
Should those unit additions in `load_camels.py` be on every simulation load data script?

WillJRoper:
Arguably yes. I can do that on this or a separate PR if you like? We should probably want units passed on everything by default because otherwise the default unit system is assumed.

christopherlovell:
Let's do it here. I can add it

christopherlovell (approved):

PR #443, opened 2023-11-23
Parametric morphology improvements

stephenmwilkins:
A few improvements to `parametric.morphology`, including:
- changed doc string to match adopted format
- added option to provide `r_eff` with units to `Sersic2D` class for automatic calculations.
- added a `PointSource` class for modelling nuclear/unresolved components
- added a nuclear component to `generate_composite_galaxy` docs as a placeholder until we can add a blackhole component.
- discovered a hitherto unknown bug where specific choices of `fov` or `npix` cause weirdness.

## Checklist


src/synthesizer/parametric/morphology.py
```diff
@@ -46,30 +51,64 @@ class Sersic2D(MorphologyBase):
...
+                Ellipticity.
+            theta (float)
+                Theta, the rotation angle.
+            cosmo (astro.cosmology)
+                astropy cosmology object.
+            redshift (float)
+                Redshift.
+
+        """
+        self.r_eff_mas = None
```
WillJRoper, line 85: You've left the arguments for these but then explicitly set them to None here. Probably worth just fully removing the specific arguments.

src/synthesizer/parametric/morphology.py
```diff
@@ -46,30 +51,64 @@ class Sersic2D(MorphologyBase):
...
+            if r_eff.units.dimensions == length:
+                self.r_eff_kpc = r_eff.to('kpc').value
+            elif r_eff.units.dimensions == angle:
+                self.r_eff_mas = r_eff.to('mas').value
+            else:
+                raise exceptions.IncorrectUnits(
+                    "The units of r_eff must have length or angle dimensions"
+                    )
+            self.r_eff = r_eff
+        elif r_eff_mas:
```
WillJRoper, line 99: I see that here the argument is used. Still, I think the singular argument is good enough.

src/synthesizer/parametric/morphology.py
```diff
@@ -182,3 +219,104 @@ def compute_density_grid(self, xx, yy, units=kpc):
...
+        offset=np.array([0., 0.])*kpc,
+        cosmo=None,
+        redshift=None,
+    ):
+        """
+        Initialise the morphology.
+        
+        Arguments
+            offset (unyt)
+                The [x,y] offset.
```
WillJRoper, line 242: Offset from what?

WillJRoper (changes requested):

src/synthesizer/parametric/morphology.py
stephenmwilkins: I've removed the arguments but kept the attributes for now since it's useful to record both.

src/synthesizer/parametric/morphology.py
stephenmwilkins: Updated doc string - offset from the centre of the image.

src/synthesizer/parametric/morphology.py
WillJRoper: Agreed, that's how `parametric.Stars` works with the single argument-multiple method.

WillJRoper (approved):

PR #444, opened 2023-11-24
Fix and complete sanity check of particle spectra derivation

### LINKED ISSUE #278: Provide optional grid assignment schemes (CIC, NGP, etc.)
Whilst it may make only a small difference to production results, it would be useful to have some alternative grid assignment schemes available (e.g. Nearest grid point, cloud in cell) to assess the impact. It may also be important where arbitrary axes are in either linear or log space.

The solution would be to create these different schemes in the C extensions (e.g. [here](https://github.com/flaresimulations/DensityGridder/blob/12c1b555952a104018a09716f87264d2ce179cb3/gridder.c#L305)) and have keywords to switch between them.

WillJRoper:
This PR addresses the issues observed when comparing parametric and particle derived spectra. 

The code surrounding this process was somewhat convoluted due to the parts that needed to handle arbitrary dimensions. This complexity had lots of scope to introduce mistakes and one had been introduced in the assignment of grid indices. The cause of this issue was a logical expression which should have been `>=` but was just `>`. 

This PR has not only fixed the code and simplified it dramatically it:
- Removes the recursive method in favour of a clearer iterative method that still works for arbitrarily many directions.
- Closes #278, by the introduction of a `grid_assignment_method` argument which can either take `"cic"/"CIC"` to use the old "interpolation" like method, or `"ngp"/"NGP"` to use a new nearest grid point method.
- Introduces `**kwargs` in all spectra method arguments to enable the passing of keywords to `generate_lnu`/`generate_particle_lnu`. This addresses an issue where most spectra methods couldn't turn off the grid bounds check.
- Introduces methods (and C extensions) to calculate a binned SFZH (i.e. the grid weights) for particle based `Stars` objects, analogous to the parametric SFZH. This makes sanity checks easier since both the parametric and particle `Stars` objects can now be compared on common ground.
- Adds an example test comparing a single particle SFZH and an instantaneous SFZH.
- Introduces an example comparing the grid assignment methods.

In the process of checking this works I checked:
- Single stars match between the particle and parametric methods.
- A single star particle when sampling exactly on a grid point behaves as expected.
- A single star particle when sampling between grid points behaves as expected, with the correct assignment of mass to grid points.
- The two assignment schemes give comparable results.

Here is a comparison of single star SFZH histories with the fixed code (CIC):

Fixed_CIC

Here is a comparison of single star SFZH histories with the fixed code (NGP):

Fixed_NGP

And here are the spectra (exactly overlaid on each other):

sucess

Finally here is the comparison between assignment methods:

grid_comp

All of these plots can be produced by the examples.

## Issue Type

- Bug
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/particle/stars.py
+++ b/src/synthesizer/particle/stars.py
@@ -264,7 +264,14 @@ class Stars(Particles, StarsComponent):
 
         return pstr
 
-    def _prepare_sed_args(self, grid, fesc, spectra_type, mask=None):
+    def _prepare_sed_args(
+        self,
+        grid,
+        fesc,
+        spectra_type,
+        mask,
+        grid_assignment_method,
+    ):
         """
         A method to prepare the arguments for SED computation with the C
         functions.
@@ -280,6 +287,15 @@ class Stars(Particles, StarsComponent):
             mask (bool)
                 A mask to be applied to the stars. Spectra will only be computed
                 and returned for stars with True in the mask.
+            grid_assignment_method (string)
+                The type of method used to assign particles to a SPS grid
+                point. Allowed methods are cic (cloud in cell) or nearest
+                grid point (ngp) or there uppercase equivalents (CIC, NGP).
+                Defaults to cic.
+
+        Returns:
+            tuple
+                A tuple of all the arguments required by the C extension.
         """
 
         # Make a dummy mask if none has been passed
@@ -326,6 +342,7 @@ class Stars(Particles, StarsComponent):
             len(grid_props),
             npart,
             nlam,
+            grid_assignment_method,
         )
 
     def generate_lnu(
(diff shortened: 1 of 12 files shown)


src/synthesizer/particle/stars.py
```diff
@@ -264,7 +271,14 @@ def __str__(self):
...
         return pstr
 
-    def _prepare_sed_args(self, grid, fesc, spectra_type, mask=None):
+    def _prepare_sed_args(
+        self,
+        grid,
+        fesc,
+        spectra_type,
+        mask,
+        grid_assignment_method,
```
christopherlovell, line 280: Not sure this does default to CIC here, only in parent methods?

christopherlovell (approved):

PR #446, opened 2023-11-24
FIx self reference in apply_attenuation

christopherlovell:
The `apply_attenuation` method on Sed references a `tau_v` array on Sed, rather than the argument passed to the function, leading to an error.

This patch fixes the reference to the argument

## Issue Type
- Bug

## Checklist
- [x ] I have added docstrings to all methods


WillJRoper (approved):
Woof... that's an oversight!

WillJRoper:
No idea how this wasn't triggered by any tests!

PR #466, opened 2023-11-30
Added dust emission to PacMan model

### LINKED ISSUE #463: Rename lnu method on dust emission model as get_spectra to align with blackhole emission models
Rename `lnu` method on dust emission model as `get_spectra` to align with blackhole emission models (and other objects).

### LINKED ISSUE #464: Automatically generate dust emission if a dust emission model is provided
When calling complex models e.g. CF00, Pacman automatically calculate the dust emission spectrum if a dust emission model is provided.

stephenmwilkins:
Handful of changes:

- Dust emission models have been modified to have a renamed method `get_spectra` which in this case returns the normalised spectra as an `Sed` object.
- If you pass a `dust_emission_model` (or models) to the `get_spectra_pacman` it now generates the dust emission spectra, including for both young and old components individually. For young it also computes both the birth cloud and ISM dust emission.
- Updated `synthesiser.blackholes` which uses a dust emission model to model the torus.
- Added diffuse dust emission and attenuation to `components.blackholes.get_spectra()` so we can now easily model the diffuse attenuation. 
- Modified relevant doc notebooks.

Closes #463, closes #464 

## Issue Type

- Bug
- Document
- [x] Enhancement

## Checklist


WillJRoper:
Since they are here: Can the diffuse dust be split into a separate method (`get_spectra_attenuated`) like the stellar methods instead of being in a catch-all method? We need to get everything matching up across all galaxy components.

Really `get_spectra` should be `get_spectra_incident` or similar to be consistent with the rest of the package.

stephenmwilkins:
> Since they are here: Can the diffuse dust be split into a separate method (`get_spectra_attenuated`) like the stellar methods instead of being in a catch-all method? We need to get everything matching up across all galaxy components.
> 
> Really `get_spectra` should be ~`get_spectra_incident`~ `get_sepctra_intrinsic` or similar to be consistent with the rest of the package.

Done. Split into `get_spectra_intrinsic` and `get_spectra_attenuated`; However, have left get_spectra as an alias for `get_spectra_attenuated` since I think it makes more sense for now.

WillJRoper (approved):

PR #471, opened 2023-11-30
Add the ability to plot luminosity or flux rather than density

### LINKED ISSUE #465: Would be nice to be able to plot luminosity as well as lnu
Plotting `luminosity = nu*lnu` is sometimes more convenient than plotting `lnu`. Would be nice to be able to choose, while plotting `lnu` by default.

WillJRoper:
This PR enables the passing of the quantity to plot to `plot_spectra` methods enabling the plotting of `lnu`, `llam`, `luminosity`, `fnu`, `flam`, and `flux` (which has been added as a property method). Error handling has been included to ensure the argument is valid before a less obvious error is triggered. 

This cleans up a lot of the logic in the base functions while improving flexibility. 

A demonstration of this has been added to `generate_sed.ipynb`.

This also fixes an issue where infinities would break the automated y limit calculation.

Closes #465 

## Issue Type

- Enhancement

## Checklist


WillJRoper:
Can do, the units are handled automatically by the unit system or the user can pass units explicitly so that is dealt with.

christopherlovell:
I agree that `plot_density` is ambiguous, perhaps rename to `plot_spectral_density` if that is the approach we are going for. 

An optional argument to choose the property to plot, as suggested by Steve, would be nice, but risks making this method too specialised; most users are going to want more control over their plot anyway, this should just be a quick method for eyeballing a particular spatial property

stephenmwilkins:
I think being able to specify which quantity (with `lnu` by default) isn't particularly specialised. It's a really small change right? Like I said, sometimes `luminosity` makes more sense, but most times `lnu` is fine. All we need is a single `plot_spectra` function which offers this flexibility.

WillJRoper:
But the user can do this easily outside these methods? These are for a quick look, not a publication ready plot.

WillJRoper:
Although, in hindsight... taking the property to plot is 
a) less work to implement.
b) more flexible if people really do want to use these functions.
c) maintains localisation of spectra plotting to a single function for observed and rest frame.
d) simplifies the code with less branching necessary.

I'm sold I think. Other methods lead to way more boilerplate and complications in the backend.

WillJRoper:
Cool, I'll get that in now.

WillJRoper:
Done, much more elegant!

stephenmwilkins (approved):

PR #473, opened 2023-11-30
Moved transmission and attenuation functions to sed.py

### LINKED ISSUE #392: Transmission, A, Al, AV and A1500 shouldn't be attached to a galaxy
These methods are more generically useful than purely for use with a `Galaxy`. They should either:

- Be functions in `sed.py` that take two `Sed` objects to.
- Be methods on an `Sed` that take another `Sed` as an argument.

My preference (but not a strong one) is for the former.

WillJRoper:
This PR moves the following methods from a `Galaxy` to be standalone functions in `sed.py`:
- `transmission`
- `A` 
- `Al`
- `Av`
- `A1500`

These functions now have names following the `get_` style convention and snake_case style. They take an intrinsic and an attenuated `Sed` as arguments and any that take a wavelength take one with units to enable automatic conversion and consistency regardless of the passed units.

This is all demonstrated in `sed.ipynb` in the docs.

Closes #392 

## Issue Type

- Enhancement

## Checklist
- [x] I have made corresponding changes to the documentation (doesn't feature yet)

### DIFF
--- a/docs/source/sed.ipynb
+++ b/docs/source/sed.ipynb
@@ -781,7 +781,7 @@
    "outputs": [],
    "source": [
     "from unyt import um, angstrom\n",
-    "from synthesizer.sed import get_attenuation, get_attenuation_lam, get_attenuation_1500, get_attenuation_5500\n",
+    "from synthesizer.sed import get_attenuation, get_attenuation_at_lam, get_attenuation_at_1500, get_attenuation_at_5500\n",
     "\n",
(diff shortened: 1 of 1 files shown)


WillJRoper:
I originally had the at in there but removed it. I'll add it back in

WillJRoper:
- Done.
- I think that's a bit redundant, that is what `get_transmission(...)` does.
- Done.

stephenmwilkins (approved):

PR #476, opened 2023-12-01
Enabling per particle escape fractions

### LINKED ISSUE #124: fesc could be an array.
Should we assume this as standard and simply make an fesc array if a single value is passed? My vote is yes. Its a small temporary memory footprint increase.

WillJRoper:
Although not a common use case it would be nice to be able to give star particles particle specific escape fractions (i.e. based on age or metallicity). This PR enables this as a possibility.

- `fesc` can be a single value. If it is an array full of that value is made to pass to the C extension. This is short lived enough and only 32 bit so has a minimal effect on footprint.
- `fesc` can also be passed as any array.
- Since the integrated spectra extension limits lam loops by calculating the weight grid first and then applying it to spectra the escape fraction is combined into the weights, this is all hidden in the backend but this approach is the only way to maintain this optimisation and enable the `fesc` array. It works but is worth noting.
- Closes #124 

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/extensions/integrated_spectra.c
+++ b/src/synthesizer/extensions/integrated_spectra.c
@@ -158,7 +158,7 @@ PyObject *compute_integrated_sed(PyObject *self, PyObject *args) {
     for (int ilam = 0; ilam < nlam; ilam++) {
 
       /* Add the contribution to this wavelength. */
-      /* fesc is already included in the weight. */
+      /* fesc is already included in the weight */
       spectra[ilam] += grid_spectra[spectra_ind + ilam] * weight;
     }
   }
--- a/src/synthesizer/extensions/integrated_spectra.c
+++ b/src/synthesizer/extensions/integrated_spectra.c
@@ -37,14 +37,14 @@ PyObject *compute_integrated_sed(PyObject *self, PyObject *args) {
 
   const int ndim;
   const int npart, nlam;
-  const double fesc;
   const PyObject *grid_tuple, *part_tuple;
   const PyArrayObject *np_grid_spectra;
+  const PyArrayObject *np_fesc;
   const PyArrayObject *np_part_mass, *np_ndims;
   const char *method;
 
-  if (!PyArg_ParseTuple(args, "OOOOdOiiis", &np_grid_spectra, &grid_tuple,
-                        &part_tuple, &np_part_mass, &fesc, &np_ndims, &ndim,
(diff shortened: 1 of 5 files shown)


aswinpvijayan:
Looks good.
Can anyone check again if this line is correct: https://github.com/flaresimulations/synthesizer/blob/ab4e2b1a0d1f7b48d3a6103c3e1d2ee6dfcdb2a6/src/synthesizer/particle/stars.py#L755
Shouldn't the nebular continuum be corrected for fesc? @stephenmwilkins

stephenmwilkins:
> Looks good. Can anyone check again if this line is correct:
> 
> https://github.com/flaresimulations/synthesizer/blob/ab4e2b1a0d1f7b48d3a6103c3e1d2ee6dfcdb2a6/src/synthesizer/particle/stars.py#L755
> 
> 
> Shouldn't the nebular continuum be corrected for fesc? @stephenmwilkins

Short answer is no, long answer is explained by the bit commented out after the line. `continuum` should be the sum of `nebular_continuum` (which should have an (`1-fesc`) factor and  `stellar_continuum` that won't. Since `stellar_continuum` is generally >> `nebular_continuum` it would normally be fine. But we should do it properly and use both separate continuum elements. Not sure why we didn't in the first place.

WillJRoper:
@aswinpvijayan can you make an issue for this outside this PR?

stephenmwilkins:
> > Looks good. Can anyone check again if this line is correct:
> > https://github.com/flaresimulations/synthesizer/blob/ab4e2b1a0d1f7b48d3a6103c3e1d2ee6dfcdb2a6/src/synthesizer/particle/stars.py#L755
> > 
> > Shouldn't the nebular continuum be corrected for fesc? @stephenmwilkins
> 
> Short answer is no, long answer is explained by the bit commented out after the line. `continuum` should be the sum of `nebular_continuum` (which should have an (`1-fesc`) factor and `stellar_continuum` that won't. Since `stellar_continuum` is generally >> `nebular_continuum` it would normally be fine. But we should do it properly and use both separate continuum elements. Not sure why we didn't in the first place.

Actually, this should be `incident_continuum` not `stellar_continuum`. I can't remember if these are actually recorded in the grids separately though.

stephenmwilkins (approved):
I'm going to assume you've done all the c stuff correctly.

PR #479, opened 2023-12-04
Fixing the age masking for spectra creation

WillJRoper:
The docs stated the `young` and `old` flags should be passed in years while the code stated it should be passed in Myrs. This PR fixes the docs but also:

- Introduces a check for whether the `young` and `old` flags exceed the age of the universe after being assumed to be in Myrs for safety when passed without units.
- Changes the default `young` and `old` values to `None` since this makes more sense than a boolean which is only ever `False`, a user may think to pass `True` which is undefined behaviour. (The docs stated they were `None` anyway.)
- Modifies the doc strings to state that a `unyt_quantity` can be passed.

The CAMELS test data in the docs doesn't have any young stars in it, so the thresholds have been changed purely for demonstration purposes. A note of this fact has been included in the docs.

## Issue Type

- Bug

## Checklist

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -107,10 +107,10 @@ class StarsComponent:
             fesc_LyA (float)
                 Fraction of Lyman-alpha emission that can escape unimpeded
                 by the ISM/IGM.
-            young (bool, float):
+            young (bool/float/unyt_quantity):
                 If not False, specifies age in Myr at which to filter
                 for young star particles.
-            old (bool, float):
+            old (bool/float/unyt_quantity):
                 If not False, specifies age in Myr at which to filter
                 for old star particles.
             kwargs
@@ -159,10 +159,10 @@ class StarsComponent:
         Args:
             grid (obj):
                 Spectral grid object.
-            young (bool, float):
+            young (bool/float/unyt_quantity):
                 If not False, specifies age in Myr at which to filter
                 for young star particles.
-            old (bool, float):
+            old (bool/float/unyt_quantity):
                 If not False, specifies age in Myr at which to filter
                 for old star particles.
             label (string)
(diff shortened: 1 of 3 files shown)


src/synthesizer/components/stellar.py
```diff
@@ -1223,32 +1221,46 @@ def get_line_screen(
...
 
-        if old:
+        if old is not None:
             if isinstance(old, (unyt_quantity)):
                 old = old.to("Myr")
             else:
                 old *= Myr
 
+        # Ensure ages aren't larger than the age of the universe
+        if young is not None and young > 13.8 * Gyr:
```
christopherlovell, line 1251: So we are only allowed to use synthesizer for generating spectra in a fiducial Lambda-CDM Universe? This is going to make it difficult to use for, e.g. CAMELS :)

christopherlovell (changes requested):

src/synthesizer/components/stellar.py
christopherlovell: You're gonna have to think of a better way, as this kind of hard check isn't going to work. Not sure what you mean by 'magnitudes too old'.

WillJRoper: That or we enforce units which would be safer

WillJRoper: For what it's worth: I don't think not having a check here is an option. By having it check for 50 Gyrs as the limit this won't exclude any meaningful cosmology but it will catch any unit issues. We've had the issue this fixes fly under the radar for ages.

WillJRoper: Ok, I'll enforce units. I prefer that anyway, I thought you would be against it 😂

christopherlovell (approved):

PR #482, opened 2023-12-05
Some clean up and consistency improvements to Blackholes/BlackHole

WillJRoper:
This PR implements some clean up and consistency changes to the black hole component:

- The parametric `BlackHoles` class has been renamed to `BlackHole` since it is by definition singular and this difference can help should we ever be checking if a black hole component is parametric or particle-based with `isinstance`.
- Removed unused imports.
- Implements a `value_to_array` helper function to wrap singular values in arrays.
- Handling singular values being passed to `particle.BlackHoles` using above helper.
- Added smoothing lengths to `particle.BlackHoles`.
- Fixed `Template.get_spectra`, it didn't use the passed bolometric_luminosity for scaling.
- Emission models are now in `blackhole_emission_models`.
- Emission models retain their functionality to get spectra using the previously implemented NGP method but particle and parametric `BlackHole/s` objects now use the same workflow used in `Stars` objects with a generalised `generate_lnu` method. For consistency, the grid look up uses the same C extension mechanism (but of course the speed is immaterial for black holes).
- The above means both CIC and NGP methods can be used for the grid look up.
- `particle.Blackholes` can now produce spectra for individual particles with `get_particle_spectra_*` methods like `Stars` can.
- `Template` emission models can now be used with `particle/parametric.BlackHoles/BlackHole`.
- Blackhole spectra methods (not on the emission models) now return `Sed`s and store `Sed`s in a `dict` on the component.
- Particle spectra are now (N, Nlam) `Sed` objects also stored in a `dict` on the component (`particle_spectra`).
- Removed `bol_luminosity` attributes in favour of the `bolometric_luminosity` attribute introduced with the AGN PR.
- Emission model parameters are also stored as attributes not just dictionaries, while the dictionaries still exist these are much more opaque and lead to more boilerplate in the methods that use them.
- `fixed_parameters` now contains only those parameters explicitly set by the user. Previously the fact all arguments had defaults meant all arguments were considered fixed. These defaults are still used if an argument is not provided but if a blackhole object defines these attributes only the truly fixed ones will be honoured.
- Black hole objects now take `**kwargs` which can be used to set attributes on the black holes and thus override any default parameters of the emission models (those not explicitly fixed by the user).
- Implemented `plot_spectra` helpers as for `Stars` and used it in the docs.
- Parametric blackholes docs are now included in the `.rst` (they were previously missing).
- Added a demonstration of particle spectra to `particle_blackholes.ipynb`.
- Now using explicit attributes in the emission models to improve readability and transparency. This does lead to some hoop jumping but now it is clear to anyone reading the code what parameters are needed for what.
- Improved the description of fixed and variable parameters in the docs with an explicit example.
- Avoids possible bug in `UnifiedAGN.get_spectra` where passing unyt masses led to nonsense kg masses by temporarily using a `Quantity`.
- `sed.measure_bolometric_luminosity` now handles multidimensional SEDs properly, previously the reversing of `lnu` scrambled the array producing nonsense results.
- Fixed bug where `PointSource` was handed an offset without units and then defined an `offset` attribute that was unused in the `density_grid` calculation.
- Now imaging is possible for stars or black holes independently or combined (for both particle and parametric).
- Parametric imaging of black holes is demonstrated in the docs.
- Included sanity check example comparing particle, parametric, and emission model spectra for a single black hole.

This PR became a lot larger than initially intended because in the act of sanity-checking things I found several unrelated bugs (such as the bolometric luminosity and morphology bugs). Everything now works and gives matching results (as demonstrated in the single bh example). As far as I am aware the implementation is consistent with the rest of Synthesizer.

Importantly all results are identical to those produced on main for the black holes, modulo the particle-specific spectra which were previously wrong due to bolometric luminosity issues. The shape of these spectra is identical to the previous result but now the normalisation is correct.

## Issue Type

- Enhancement

## Checklist


christopherlovell:
I think differentiating between particle and parametric based on pluralisation is going to be more confusing for the user, and you also seem to describe above how users should check the version they are using (`isinstance`). No strong feelings though

stephenmwilkins (approved):
Looks good to me, I've mostly focussed on checking the notebooks. I have one lingering concern but that's pretty separate from this PR.

PR #485, opened 2023-12-06
Updated plot_equivalent_width and sed.py to accept multiple SEDs

ConnorSF:
## Issue Type

- Enhancement

## Checklist


src/synthesizer/base_galaxy.py
```diff
@@ -105,6 +105,63 @@ def get_spectra_dust(self, emissionmodel):
 
         return sed
 
+    def transmission(self):
```
WillJRoper, line 108: Sorry @ConnorSF this all looks fine but you've added back in all these functions which were moved to be stand alone and better named functions in `sed.py`. Can you remove these and use those if you need them.

WillJRoper: These being `transmission`, `Al`, `A`, `Av`, and `A1500`

WillJRoper (changes requested):
Sorry

src/synthesizer/base_galaxy.py
ConnorSF: No problem, will remove asap

WillJRoper (approved):

PR #489, opened 2023-12-15
Introduces a method for combining component spectra on a `Galaxy`

WillJRoper:
This PR is dependant on #482.

This introduces a method (`get_spectra_combined`) for combining spectra from components onto galaxy level spectra. 

- First the components are checked for instances of "intrinsic", "emergent", and "total" spectra. 
- Any of these keys on more than one component are combined into a new `Sed` in `BaseGalaxy.spectra`. This means, for instance, that if only the stars component has "emergent" spectra then no combined spectra will be produced. This could instead be flipped on its head such that any of the 3 spectra that can be combined are included as long as 1 or more instances exist. I'm not sure which is better, both could be confusing. In my mind, if only 1 instance exists it can't be "combined".
- Introduced a demo of this method in the `generate_active_galaxy` docs.
- Seemingly separately (but required for `sum([*spectra])` which is used in this PR), `Sed.__radd__` has been introduced to overload "reflected" addition.

## Issue Type

- Enhancement

## Checklist


src/synthesizer/extensions/particle_spectra.c
```diff
@@ -61,7 +61,7 @@ void spectra_loop_cic(const double **grid_props, const double **part_props,
 
       /* Use the grid edge. */
       part_cell = 0;
-      frac = 1;
```
christopherlovell, line 64: Why are we not using the CIC and NGP methods in `weights.h` here? All the code appears to be identical up to the weight summation

christopherlovell (changes requested):

src/synthesizer/extensions/particle_spectra.c
WillJRoper: You can't use that code here. This is per particle so you can't make a single grid of weights. If you made a grid of weights per particle the memory footprint would be prohibitive. That's why this is bespoke and loops over the spectra for each particle.

PR #491, opened 2023-12-21
Introduced a PhotometryCollection

### LINKED ISSUE #384: A `PhotometryCollection` would enable units and nice functionality
As much as I want to avoid introducing more and more classes... I've been convinced a photometry container class would be nice both for associating units with results and for adding helper functionality.

There may be some gotchas with how exactly to handle the creation of `PhotometryCollection`s but it will probably just require instantiating one and adding photometry after instantiation rather than at instantiation. This is because Photometry is calculated on a per `Filter` basis and not at a `FilterCollection` level, and I don't think that should change!

WillJRoper:
Closes #384 by introducing a `PhotometryCollection`.

This new class:
- Is only every instantiated by `Sed.get_broadband_luminosities` and `Sed.get_broadband_fluxes` and acts as a container for the photometry data.
- Is stored on a `Sed` where dictionaries previously were, under `Sed.broadband_luminosities` and `Sed.broadband_fluxes`.
- Provides some quick plotting functionality to look at filter curves and their resulting photometry.
- Provides a nice summary of photometry:
```
==================== REST FRAME PHOTOMETRY =====================
----------------------------------------------------------------
| U (λ = 3.64e+03 Å) | V (λ = 5.51e+03 Å) | J (λ = 1.22e+04 Å) |
|--------------------|--------------------|--------------------|
|   6.60e+27 erg/s   |   6.37e+27 erg/s   |   8.88e+27 erg/s   |
----------------------------------------------------------------
```
- Overloads `__getitem__` such that a `PhotometryCollection` can behave exactly like a dictionary (e.g. `sed.broadband_photometry["U"]` will return the U band photometry with units).
- Contains `rest_photometry` and `obs_photometry` `Quantity` attributes containing the an array of photometry depending on the frame. Previously units were hard coded which could violate a modified unit system. This is a necessity because the units of luminosity and flux differ. 
- Contains a `photometry` attribute for the UI to get an array of photometry with units. Thus, the user doesn't need to worry about the frame if they don't want to. This will ensure the unit system is respected.

This PR also introduces a demonstration of this in generate_sed.ipynb.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/photometry.py
+++ b/src/synthesizer/photometry.py
@@ -109,26 +109,93 @@ class PhotometryCollection:
             ax = fig.add_axes((left, bottom, width, height))
 
             # Set the scale to log log
-            ax.loglog()
+            ax.semilogy()
 
         # Add a filter axis
         filter_ax = ax.twinx()
         filter_ax.set_ylim(0, None)
 
         # PLot each filter curve
+        max_t = 0
         for f in self.filters:
             filter_ax.plot(f.lam, f.t)
+            if np.max(f.t) > max_t:
+                max_t = np.max(f.t)
+
+        # Get the photometry
+        photometry = self.rest_photometry if self.rest_frame else self.obs_photometry
 
         # Plot the photometry
-        for f, phot in zip(self.filters, self._photometry):
+        for f, phot in zip(self.filters, photometry.value):
             pivwv = f.pivwv()
             fwhm = f.fwhm()
-            ax.errorbar(pivwv, phot, marker=marker, xerr=fwhm, linestyle=None)
+            ax.errorbar(
+                pivwv,
+                phot,
+                marker=marker,
+                xerr=fwhm,
+                linestyle=None,
+                capsize=3,
+            )
+
+        # Do we not have y limtis?
+        if len(ylimits) == 0:
+            max_phot = np.max(photometry)
+            ylimits = (
+                10 ** (np.log10(max_phot) - 5),
+                10 ** (np.log10(max_phot) * 1.1),
+            )
+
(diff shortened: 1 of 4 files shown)


src/synthesizer/photometry.py
```diff
@@ -0,0 +1,334 @@
...
+        else:
+            title = f"{'= OBSERVED PHOTOMETRY ='.center(tot_width, '=')}"
+
+        # Combine everything into the final table
+        table = (
+            f"{title}\n-{sep.replace('|', '-')}-\n|{head}|"
+            f"\n|{sep}|\n|{phot}|\n-{sep.replace('|', '-')}-"
+        )
+
+        return table
```
christopherlovell, line 190: maybe I'm misunderstanding how this works, but surely a two column table, with name and details, would be much easier to read, especially where there are >10 filters in a collection?

christopherlovell (changes requested):

christopherlovell:
Have reviewed, looks fine, would just say that this seems like an awful lot of code to maintain just to enable plotting and string printing. Not sure I understand the need for 'photometry units' either

WillJRoper:
> Have reviewed, looks fine, would just say that this seems like an awful lot of code to maintain just to enable plotting and string printing. Not sure I understand the need for 'photometry units' either

It 100% would 😂 don't know why I chose a scalable row 🤦‍♂️😅 will fix!

src/synthesizer/photometry.py
WillJRoper: Done, much better!

christopherlovell (approved):

PR #499, opened 2024-01-08
Introduces a better name for photometry

### LINKED ISSUE #490: Is `broadband_luminosity/flux` the best naming convention?
We call photometry `broadband_*` which has just struck me as odd. You can use a narrowband filter and have it stored in a "broadband" luminosity attribute. 

Solution:

Rename these variables to something more generic like `rest_photometry` and `obs_photometry`.

WillJRoper:
DWISOTT:

`broadband_luminosities` -> `photo_luminosities`
`broadband_fluxes` -> `photo_fluxes`

Closes #490.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/docs/source/cosmo/cosmo_example.ipynb
+++ b/docs/source/cosmo/cosmo_example.ipynb
@@ -194,7 +194,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "_UVJ = spec.get_broadband_fluxes(fc)\n",
+    "_UVJ = spec.get_obs_photometry(fc)\n",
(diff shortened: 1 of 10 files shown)


christopherlovell:
rest frame flux does not equal luminosity. I don't think this is a wise change 

I would prefer `photo_luminosities` and `photo_fluxes`, where the latter can be in rest frame or observed frame depending on what the user has specified

WillJRoper:
Done

christopherlovell (approved):

PR #509, opened 2024-01-16
IR template model, DL07 model added

aswinpvijayan:
This PR adds in a dust emission model class to hold IR templates based on the DL07 (https://ui.adsabs.harvard.edu/abs/2007ApJ...657..810D). The updated 2014 grids are used here. This model does not use the emission base as that is a temperature dependent model, while this is a dust mass (and dust luminosity, in our case using energy balance) based model giving the contribution of dust emission from PDRs (hot dust, contribution to the mid-IR) and diffuse dust. 

Within components/stellar.py, the emission model components have been assigned as follows:
pdr -> young_dust
diffuse -> old_dust

I have added the grid hdf5 file for the spectrum as downloaded into the test_grid folder for now.

## Issue Type
- Enhancement

## Checklist


src/synthesizer/utils.py
```diff
@@ -102,6 +102,134 @@ def rebin_1d(arr, resample_factor, func=np.sum):
     return func(brr, axis=1)
 
 
+def process_dl07_to_hdf5(grid_name='MW3.1', grid_loc='./',
```
WillJRoper, line 105: This feels oddly placed. It should probably be in the grids repo.

WillJRoper: Could add a `dust` module for it. Since you're effectively making a grid here.

WillJRoper (changes requested):
Looks great overall. My only thought (as commented) is that the conversion from the downloaded file to a hdf5 file should probably be in the grids repo not in utils. I'm not dead set on that though so feel free to argue against that. What do you think @christopherlovell?

src/synthesizer/utils.py
aswinpvijayan: No problem. I can remove it from here and add it in the grids repo.

WillJRoper (approved):

PR #513, opened 2024-01-25
Introduced a data downloading tool

WillJRoper:
With data now on a server we can provide a more polished method for downloading any of our data requirements. 

This PR introduces a download tool in `src/synthesizer/download_data.py`. Although this can be used by running this script, an entry point has been added enabling the following on the command line:

```
synthesizer-download --test-grids -d tests/test_grid/
```

The currently implemented options are best summarised by the help output.

```
➜  synthesizer-download --help
usage: synthesizer-download [-h] [--test-grids] [--dust-grid] [--camels-data] [--camels-snap CAMELS_SNAP] [--camels-lh CAMELS_LH] [--all] --destination DESTINATION

Download datasets for synthesizer

options:
  -h, --help            show this help message and exit
  --test-grids, -t      Download the test data for synthesizer
  --dust-grid, -D       Download the dust grid for the Drain & Li (2007) model
  --camels-data, -c     Download the CAMELS dataset
  --camels-snap CAMELS_SNAP
                        Which snapshot should be downloaded? (Default: 031)
  --camels-lh CAMELS_LH
                        Which LH variant should be downloaded? (Default: 1)
  --all, -A             Download all available data including test data, camels simulation data, and the dust grid
  --destination DESTINATION, -d DESTINATION
                        The path to the destination directory
```

Note that the CAMELS data is not on the server yet. For this, an `UnimplementedFunctionality` error is triggered for now stating this. I should have this uploaded soon, but we need a discussion about what data we have where.

It would be good to also add the SAM data @louiseseeyave added to this along with any other data dependencies we have. Any data that will always be needed can be put under a `--setup` flag and we can include this in the installation instructions. That way we can remove all data from `src`!

I'll add docs for this once we have fully implemented it.

## Issue Type

- Enhancement

## Checklist


christopherlovell:
What is the purpose of allowing users to download the raw CAMELS data? The processed data, of only 10 galaxies, should be downloadable, but we don't necessarily want / need to have an API point for the whole CAMELS LH set

WillJRoper:
> What is the purpose of allowing users to download the raw CAMELS data? The processed data, of only 10 galaxies, should be downloadable, but we don't necessarily want / need to have an API point for the whole CAMELS LH set

Agreed, I was just following the download scripts that were already in Synthesizer and converting them over. This PR isn't about changing that behaviour it was simply porting download model. If we don't want the user to have that capability and we just want to give them a cut-down download we can upload that data and enable that. I think that's separate from this PR, currently, the CAMELS data isn't on the server anyway. This was simply a skeleton for what download functionality we may want including everything that was a standalone script already in Synthesizer.

PR #516, opened 2024-01-31
Complete rewrite of the imaging module

### LINKED ISSUE #383: `Image` would be clearer if named `ImageCollection`
Currently the imaging module has `Image`, `ParametricImage` and `ParticleImage` classes. However, all of these can contain multiple images or singular images. This is handled via attribute dicts with pluralised names and attributes with singular names. 

This would be much cleaner (and fit with other naming schemes if these were "`Collections`". 

This could also enable `ImageCollection[filter].data` syntax and allow for units to be associated to images cleanly with `Quantity` objects. Strictly this could require splitting off `Image` into a singular class like `Filter`. This is yet another object but one the user should never interact with directly so probably ok.

WillJRoper:
The imaging module needed some long overdue TLC. The PR updates the imaging module to reflect the landscape of the rest of the package. To summarise:

- Images are now contained in an `Image` object which contains the methods for generating a single image either as a smoothed image or a histogram. It also holds the units of the image which aren't always the same (e.g. maps vs photometry) meaning a quantity couldn't be used. It also has a plotting helper for plotting a single image.
- Groups of related images are now held in an `ImageCollection`. This contains a bunch of functionality needed for manipulating and interacting with collections of images. It also has methods for generating multiple images at once from a `PhotometryCollection`. 
- An `ImageCollection` overloads all `dict` methods meaning it behaves much like a dictionary when interacted with but also enables list-like looping over images.
- The `Scene` no longer exists. This introduces some boilerplate here and there but vastly simplifies the structure since the crazy inheritance hierarchy is gone.
- Images no longer need to be square, both the `fov` and `npix` are internally arrays. If a single value is passed for either an array of size 2 is made from that single value.
- Including PSFs is now much more intuitive. `Image`s and `ImageCollection`s have an `apply_psf/s` method which will take a PSF or multiple PSFs and return a new `Image`/`ImageCollection` containing the convolved image/s.
- The same is true for noise but the functionality has been expanded. There are 3 `apply_noise/s` methods which return a new `Image`/`ImageCollection`:
    - `apply_noise_array`  which adds a passed noise array.
    - `apply_noise_from_std` which derives a noise array from a normal distribution with a passed standard deviation.
    - `apply_noise_from_snr` which derives a noise array from a depth and SNR, either in an aperture or for a point source.
-  The docs have now been updated to reflect these changes.
- The maps have been split out into their own docs entry.
- Morphology objects now have the ability to calculate their `density_grid` from a `resolution` and `npix`.
- The `SpectralCube` has been completely rewritten to reflect the changes to the images. It now is also much simpler and follows the same model as the images (although there is no collection since that would be meaningless). 
- A C extension has been added for the unsmoothed data cube method. This was previously in pure python.
- A `SpectralCube` now takes a wavelength array and when passed an `Sed` will interpolate the spectra onto it.
- A refined animation method has been introduced to visualise `SpectralCube`s.
- `SpectralCube`s can now be added together.
- `SpectralCube`s can now be generated from a galaxy.
- Docs have been added for the `SpectralCube`.
- Units are now fully implemented throughout the imaging module with catches and handling everywhere.

To get the `SpectralCube` implementation to work I had to implement a few nice features on the `Sed` mostly around when new `Sed` objects are made:
- When adding two `Sed`s the `fnu` arrays are added if present on both and `obslam` and `obsnu` are copied over to the new `Sed`.
- When resampling an `Sed` the `fnu` array is also resampled if present and `obslam` and `obsnu` are calculated from the new `lam`.
- When summing an `Sed` if the `fnu` array us present it is also summed and `obslam` and `obsnu` are copied over to the new `Sed`.

Closes #383 

## Issue Type

- Document
- Enhancement

## Checklist


src/synthesizer/particle/galaxy.py
```diff
@@ -1486,19 +1447,134 @@ def make_ssfr_map(
...
+        lam,
+        cube_type="hist",
+        stellar_spectra=None,
+        blackhole_spectra=None,
+        kernel=None,
+        kernel_threshold=1,
+        quantity="lnu",
+    ):
+        """
+        Make a SpectralCube from an Sed.
```
christopherlovell, line 1479: shouldn't this be from a `galaxy` object?

WillJRoper: Should what be from a galaxy object? Sorry, I don't understand what you mean from the highlighted code.

WillJRoper: Clarified now.

WillJRoper:
The unit issue is fixed now so this PR passes.

WillJRoper:
> Instead of just returning the background (noise) map could that be somehow associated with the new `Image` object that's created?

It is attached to an `ImageCollection` when called from that level. The rationale: when using a single `Image` you're probably doing something non-photometric, or something extremely simple where carrying the array around is fine. 

Attaching it to the `Image` is possible. If it's implemented there then there's no point storing them on the `ImageCollection`. I have no strong feelings either way and can see pros and cons to both.

WillJRoper:
In fact I think I may conceptually prefer it 😅 there's an intuitiveness to that over having separate dicts for them on the `ImageCollection` and also means you can use the `dict` functionality to access the noise.

WillJRoper:
@stephenmwilkins I think you're right and that's a way better implementation. Now noise and weights are stored on the new `Image` object.

christopherlovell (approved):

PR #526, opened 2024-02-06
Fix photometry units and intialisation

### LINKED ISSUE #508: `PhotometryCollection` initialisation needs fixing and simplifying
- The initialisation of a `PhotometryCollection` necessarily creates `photo_luminosities` and `photo_fluxes` attributes. It would be great if these could be hidden, otherwise it is ocnfusing for the user why there is an empty attribute when you have, say, only calculated the luminosity.
- luminosities != rest frame fluxes

https://github.com/flaresimulations/synthesizer/blob/f633b73ca58fffc8a13a062280a841ac9c067c99/src/synthesizer/photometry.py#L74

christopherlovell:
Fixes a bug where photometry was not stored in spectral flux/luminosity density units, leading to issues with downstream tasks and removes the `rest_frame` flag when instantiating a `PhotometryCollection`.

- Photometry is now in spectral flux/luminosity density units.
- Whether a flux or luminosity has been passed is automatically detected.
- All rest frame vs observer frame nomenclature has been removed.
- Small fix applied to unit formatting on plot axes.
- `Sed.get_fnu` associates the `fnu` or `lnu` units to the photometry after calling `apply_filter`. This was returning `dimensionless` quantities. Which is fine when the units system isn't modified but would have led to conversion issues if the system had been modified to use differing units for photometry and spectra.

While the `PhotometryCollection` isn't explicitly documented anywhere it is demonstrated in multiple places that photometry are grabbed in examples and docs.

Closes #508 

## Issue Type
- Bug

## Checklist

### DIFF
--- a/src/synthesizer/photometry.py
+++ b/src/synthesizer/photometry.py
@@ -330,12 +330,12 @@ class PhotometryCollection:
         x_units = (
             x_units.replace("/", r"\ / \ ")
             .replace("**", "^")
-            .replace("*", "\ ")
+            .replace("*", r"\ ")
         )
         y_units = (
             y_units.replace("/", r"\ / \ ")
             .replace("**", "^")
-            .replace("*", "\ ")
+            .replace("*", r"\ ")
         )
 
         # Label the x axis
--- a/src/synthesizer/photometry.py
+++ b/src/synthesizer/photometry.py
@@ -8,7 +8,6 @@ return an instance of this class.
 """
 import numpy as np
 import matplotlib.pyplot as plt
-from unyt import unyt_array, W, m
 
 from synthesizer.units import Quantity, default_units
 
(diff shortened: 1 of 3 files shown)


WillJRoper (approved):
@christopherlovell, I can't request a review from you since you made the original PR but can you take a look at the extra changes I've implemented and flag that you are ok with them before I merge?

src/synthesizer/photometry.py
```diff
@@ -33,52 +37,51 @@ class PhotometryCollection:
...
         # Store the filter collection
         self.filters = filters
 
         # Get the filter codes
         self.filter_codes = list(kwargs.keys())
 
         # Get the photometry
-        photometry = np.array(list(kwargs.values()))
+        photometry = list(kwargs.values())
+        print(photometry, photometry[0].units.dimensions)
```
christopherlovell, line 71: superfluous print statement

PR #527, opened 2024-02-06
Fixed powers and whitespace in spectra labels

WillJRoper:
DWISOTT: automatic parsing of `unyt` units into an axis label was missing whitespace and a handler for powers.

## Issue Type

- Bug

## Checklist

### DIFF
--- a/src/synthesizer/photometry.py
+++ b/src/synthesizer/photometry.py
@@ -6,6 +6,7 @@ internal methods that calculate photometry
 (e.g. Sed.get_photo_luminosities)
 return an instance of this class.
 """
+import re
 import numpy as np
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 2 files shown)


stephenmwilkins:
Ideally we'd use the `latex_representation` of the units right? Is the issue that is forces things to be fractions instead of using powers?

WillJRoper:
This does use the latex representation. The issue was the conversion from `unyt` string to latex was ignoring the spaces in math mode and I had blindly removed multiplication before handling powers.

stephenmwilkins:
But e.g. `str(plt_spectra.units)` doesn't produce the latex representation right?

WillJRoper:
No, that's what the replacing does?

stephenmwilkins:
But why not using the `latex_representation` property built into `unyt`? 

```
>>> from unyt import erg, s, Hz, cm
>>> fnu = 1 * erg/s/Hz/cm**2
>>> print(fnu.units)
erg/(Hz*cm**2*s)
>>> print(fnu.units.latex_repr)
\frac{\rm{erg}}{\rm{Hz} \cdot \rm{cm}^{2} \cdot \rm{s}}
```

The undesirable aspect of this is that uses \frac.

WillJRoper:
😂 because I literally just learnt it exists! I'm now torn... I'd rather not use frac but my current implementation isn't very cleaver.

WillJRoper:
Yea, bear with me. I can achieve it with regex. Commit incoming.

WillJRoper:
Done. It now uses Latex representation and swaps `frac` for `\`.

stephenmwilkins (approved):

PR #532, opened 2024-02-20
Add cloudy23 functionality

stephenmwilkins:
This adds a Cloudy C23.01 functionality by adding a new module (`cloudy23`). This is largely identical to the `cloudy17` module but adds a new option for accounting for depletion which has been added in Cloudy C23.01. While some shared functions could be separated into a separate master cloudy module I don't think this is a good idea as it presupposes that future cloudy releases will also be compatible. For this reason I believe it makes sense to keep them self-contained.

These changes have been tested using synthesiser-grids (see discussions on Slack). For the same choice of parameters there appears to be little difference between cloudy23 and cloudy17 which is encouraging. What changes there are likely reflect improvements made by C23.01.

## Issue Type

- Bug
- Document
- [x] Enhancement

## Checklist


src/synthesizer/photoionisation/cloudy23.py
stephenmwilkins, line 1: This has a lot of overlap with cloudy17 and potentially some functions could be shared between the two modules. However, that may make backward compatibility issues in the future so it might be best to leave things self contained like this since we have no control over cloudy output formatting.

newmansoph:
I'm just running install_maraston13 again, then will try to run this and hopefully merge

PR #544, opened 2024-03-12
Swap integration routines

christopherlovell:
Swaps integration routines from `quad` to the trapezoid method. The former is exceedingly slow for little benefit, especially since it is often applied to linearly interpolated data anyway.

## Issue Type
- Enhancement

## Checklist


WillJRoper:
Big fan of this change. Time to `git grep quad` and change this everywhere 🥳

christopherlovell:
> @christopherlovell where do we stand with this? Do we just need to go through and make the changes or are there issues to resolve (beyond the failure)?

Needs verifying, but other than that just need to apply throughout

WillJRoper:
I've gone off the deep end (as usual) and replaced the integration routines with C extensions. In all cases, we always integrate over the final axis. This means we can throw away a lot of the complicated slowdown introduced by the more general functions `scipy` or `numpy` offer. 

The new extensions are faster and just as accurate as the old alternatives and also enable threading (because why not) but this will only pay dividends for the largest integrations. They are also generalised to any dimension which was not previously true of the methods they have been used in.

This PR now introduces the integrate module, integration extension and a test to ensure the integration routines give them same answer as `scipy` for a range of different array shapes.

WillJRoper (approved):

PR #546, opened 2024-03-14
Minor updates to lines

stephenmwilkins:
Minor updates to lines including:
- Line collection contains an ordered list of lines
- Line collection automatically identify available ratios and diagnostics
- Improved line, ratio, and diagnostic label creation
- Added ability to set abundances relative to another elements (e.g. N/O)
- Separated out element data into `abundances/element`
- Added GalacticConcordance abundance pattern

## Issue Type

- Bug
- Document
- [x] Enhancement

## Checklist


docs/source/abundances.ipynb
```diff
@@ -21,8 +21,9 @@
     "import numpy as np\n",
     "from synthesizer.abundances import (\n",
     "    Abundances,\n",
+    "    Elements,\n",
     "    abundance_scalings,\n",
-    "    solar_abundance_patterns,\n",
+    "    reference_abundance_patterns,\n",
```
WillJRoper, line 26: This is going to break `synthesizer-grids`, no?

WillJRoper: Although I do prefer the naming convention

docs/source/parametric/generate_lines.ipynb
```diff
@@ -172,7 +172,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "synthesizer-env",
+   "display_name": "venv",
```
WillJRoper, line 175: You haven't got notebook sanitisation set up by the looks of things.

src/synthesizer/abundances/elements.py
```diff
@@ -0,0 +1,137 @@
...
+        all_elements (list, string)
+            A list of all elements, functionally the concatenation of metals
+            and non-metals.
+        alpha_elements (list, string)
+            A list of the elements classified as alpha-elements.
+        name (dict, string)
+            A dictionary holding the full name of each element.
+        A (dict, float)
+            Atomic mass of each element (in amus).
+    """
```
christopherlovell, line 21: This class should have an init method

src/synthesizer/abundances/elements.py
```diff
@@ -0,0 +1,137 @@
...
+    name["Cr"] = "Chromium"
+    name["Mn"] = "Manganese"
+    name["Fe"] = "Iron"
+    name["Co"] = "Cobalt"
+    name["Ni"] = "Nickel"
+    name["Cu"] = "Copper"
+    name["Zn"] = "Zinc"
+
+    # mass of elements in amus
+    A = {}
```
christopherlovell, line 107: should use a pep8 friendly variable name here, not just `A`

src/synthesizer/abundances/reference_abundance_patterns.py
```diff
@@ -54,6 +54,56 @@ class Asplund2009:
     }
 
 
+class GalacticConcordance:
+
+    """
+    The Galactic Concordance pattern from Nicholls et al.
+    """
```
christopherlovell, line 61: This is also not a properly initialised class. I'm also confused why this is independent of the abundance_patterns submodule, shouldn't this inherit from that parent class? Each of these is an independent new class

src/synthesizer/line.py
```diff
@@ -9,30 +9,130 @@
...
+    """
 
     if isinstance(id, list):
         return ",".join(id)
     else:
         return id
 
 
+# dictionary of common line labels to use by default
+special_line_labels = {
```
christopherlovell, line 35: This should not be a global. Can it be included on a `LineCollection`?

src/synthesizer/abundances/abundance_patterns.py
```diff
@@ -19,171 +19,34 @@
...
-    A["Mn"] = 54.938
-    A["Fe"] = 55.845
-    A["Co"] = 58.933
-    A["Ni"] = 58.693
-    A["Cu"] = 63.546
-    A["Zn"] = 65.38
-
-
-class Abundances(ElementDefinitions):
+class Abundances(elements.Elements):
```
WillJRoper, line 29: What is the purpose of this inheritance? I know this was here before but I've just looked a bit deeper and you never call `super().__init__()`. This doesn't seem like something that should be inherited. Instead, it seems it should be instantiated internally and used. In fact, `Elements` should be a dataclass (if a class at all), it has no methods and just contains the element information.

src/synthesizer/line.py
```diff
@@ -9,30 +9,130 @@
...
+    'H 1 4862.69A': r'H\beta',
+    'O 3 4958.91A,O 3 5006.84A': '[OIII]4959,5007',
+    'H 1 6564.62A': r'H\alpha',
+    'O 3 5006.84A': '[OIII]5007',
+    'N 2 6583.45A': '[NII]6583',
+    'N 2 6583.45A': '[NII]6583',
+}
+
+
+def get_line_label(line_id):
```
christopherlovell, line 46: Why have these methods been moved from a `LineCollection` to independent methods?

src/synthesizer/abundances/elements.py
```diff
@@ -0,0 +1,137 @@
+
+class Elements:
```
WillJRoper, line 2: As above: not sure why this is a class and why it's inherited.

src/synthesizer/line.py
```diff
@@ -479,6 +490,12 @@ def __init__(self, id_, wavelength_, luminosity_, continuum_):
         self.continuum_lam = lnu_to_llam(self.wavelength, self.continuum)
         self.ew = self.luminosity / self.continuum_lam  # AA
 
+        # more descriptive attribute name
+        self.equivalent_width = self.ew
```
christopherlovell, line 494: Can we not just adopt this globally throughout the package?

src/synthesizer/abundances/reference_abundance_patterns.py
```diff
@@ -54,6 +54,56 @@ class Asplund2009:
     }
 
 
+class GalacticConcordance:
```
WillJRoper, line 57: Again, these should really be dataclasses.

src/synthesizer/line.py
```diff
@@ -479,6 +490,12 @@ def __init__(self, id_, wavelength_, luminosity_, continuum_):
...
+
+        # element
+        self.element = self.id.split(' ')[0]
```
christopherlovell, line 497: Perhaps `element_str` would be more descriptive?

christopherlovell (changes requested):

src/synthesizer/line.py
```diff
@@ -9,30 +9,130 @@
...
+        line_label = '+'.join(line_labels)
+
+    return line_label
+
+
+def get_ratio_label(ratio_id):
+    """
+    Get a label for a given ratio_id.
+
+    TODO: enable label creation for arbitrary ratios
```
WillJRoper, line 87: IS this a TODO? Seems like this function does that.

WillJRoper (changes requested):
Some questions and requests. Possibly a few more but lets start with these and I'll take another look afterwards.

src/synthesizer/abundances/elements.py
stephenmwilkins: Done

src/synthesizer/abundances/reference_abundance_patterns.py
stephenmwilkins: No, these just define basic information about a reference abundance pattern, e.g. a solar pattern. These would probably be better as data classes but I don't consider it a priority to convert them.

src/synthesizer/line.py
stephenmwilkins: They feel more natural outside and it gives you option of making the label even if you don't have a line collection.

docs/source/abundances.ipynb
stephenmwilkins: It's already accounted for in a synthesizer-grids branch.

src/synthesizer/abundances/abundance_patterns.py
stephenmwilkins: Yes, it (along with reference patterns) should be data class, it's just hasn't been a priority to convert it.

stephenmwilkins: What do you mean by "instantiated internally and used". All I want is access to the attributes inside and this seemed the fastest way.

src/synthesizer/abundances/elements.py
stephenmwilkins: This thing could be a dictionary of dictionaries, or even just a yaml file, but then you've got to worry about the path to it.

src/synthesizer/abundances/reference_abundance_patterns.py
WillJRoper: Then these shouldn't be classes. 

I think it is a priority. We aren't merging more "for now" code.

src/synthesizer/abundances/abundance_patterns.py
WillJRoper: This is essentially an abuse of inheritance. If you want access to the variables why not define them on the class? The inheritance class is pointless in its current guise, just over complicates everything.

By inheriting them they also can't be changed.

When I say instantiated internally I mean:

```
class Abundances:
    def __init__(self, ...., elements=elements.Elements()):
        ...
        self.elements = elements
        etc.
```

WillJRoper: "it's just hasn't been a priority to convert it" ->  well consider now is the time it becomes a priority 😅

src/synthesizer/line.py
WillJRoper: Can a LineCollection retain wrappers around these to maintain the previous behaviour?

src/synthesizer/abundances/abundance_patterns.py
WillJRoper: To be clear I like the division out into an `elements` submodule, it just needs to be a `dataclass` and not inherited.

src/synthesizer/line.py
stephenmwilkins: I'm not sure this is useful but I've added wrappers.

src/synthesizer/line.py
stephenmwilkins: Doesn't make sense in LineCollection but I've moved it into the `get_line_label function` which is the only place it's used.

src/synthesizer/abundances/elements.py
stephenmwilkins: Changed

src/synthesizer/abundances/abundance_patterns.py
stephenmwilkins: Done

src/synthesizer/abundances/abundance_scalings.py
```diff
@@ -50,11 +55,260 @@ def C(metallicity):
...
+
+        """
+
+        a = -1.732
+        b = 2.19
+        nitrogen_to_oxygen = np.log10(10**a + 10**(oxygen_to_hydrogen + b))
+
+        return nitrogen_to_oxygen
+
+    def N(self, metallicity):
```
christopherlovell, line 252: this should have a more informative name

src/synthesizer/abundances/abundance_scalings.py
```diff
@@ -50,11 +55,260 @@ def C(metallicity):
...
+
+        """
+
+        a = -0.8
+        b = 2.72
+        carbon_to_oxygen = np.log10(10**a + 10**(oxygen_to_hydrogen + b))
+
+        return carbon_to_oxygen
+
+    def C(self, metallicity):
```
christopherlovell, line 294: this should have a more informative name

src/synthesizer/abundances/elements.py
```diff
@@ -0,0 +1,142 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Elements:
+
+    """
+    A simple class containing various useful lists and dictionaries.
```
christopherlovell, line 8: This is a very uninformative docstring, could this be updated to describe what it actually contains?

WillJRoper:
A couple more comments from me:

It would be good to remove the init in the dataclasses and define them at the class level. These could then be overridden if a dictionary is passed.

i.e. 
```
@dataclass
class C:
    a: int       # 'a' has no default value
    b: int = 0   # assign a default value for 'b'
```
will automatically generate an init of the form:
```
def __init__(self, a: int, b: int = 0):
```

WillJRoper:
Do you want the user to be able to pass a custom elements class? If so the code currently doesn't allow this it only imports the elements dataclass.

src/synthesizer/abundances/abundance_patterns.py
```diff
@@ -252,28 +112,40 @@ def __init__(
...
+                self.reference = getattr(reference_abundance_patterns,
+                                         reference)
             else:
-                raise exceptions.UnrecognisedOption("""Solar abundance pattern
-                not recognised!""")
+                raise exceptions.UnrecognisedOption("""Reference abundance
+                pattern not recognised!""")
+
+        # initialise class
+        self.reference = self.reference()
```
WillJRoper, line 143: This fails if the user passes an instantiated class.

src/synthesizer/line.py
```diff
@@ -139,124 +217,115 @@ class LineRatios:
...
+
+        # get the arguments that would sort wavelength
+        sorted_arguments = np.argsort(self.wavelengths)
+
+        # sort the line_ids and wavelengths
+        self.line_ids = self.line_ids[sorted_arguments]
+        self.wavelengths = self.wavelengths[sorted_arguments]
+
+        # include line ratio and diagram definitions dataclass
+        self.lineratios = LineRatios
```
WillJRoper, line 305: I could be wrong but I still think a dataclass needs to be called (instantiated)?

WillJRoper:
Looking at the code locally (I was fixing up some of the simpler issues to get this in) I've noticed the vast majority of the doc strings are incomplete across the abundances module. The classes/dataclasses are either missing Attribute descriptions or completely missing a doc string. I don't fully understand everything here so I can't write these. They need some detail to describe what everything is.

stephenmwilkins:
> This currently fails in the docs because renaming `N` to nitrogen (and the corresponding carbon function) breaks some functionality somewhere. I can't quite disentangle it because there's a lot of reliance on attribute/method and key consistency. The methods do need to be more descriptive than single letters so something has to be addressed here.
> 
> Maybe best to use a symbol to word lookup table. Either way, it's more in-depth to disentangle than I have time for so leaving this to you @stephenmwilkins. Not much to address, just the doc-strings, and the unresolved comments and then it can go in.

I've fixed it so we use e.g. "nitrogen" instead of "N" for method names.

src/synthesizer/abundances/abundance_patterns.py
```diff
@@ -19,171 +16,34 @@
...
             0.0 (no alpha-enhancement).
         abundances (dict, float/str)
             A dictionary containing the abundances for specific elements or
             functions to calculate them for the specified metallicity. Optional
             initialisation argument. Defaults to None.
-        solar (object)
-            Solar abundance pattern. Optional initialisation argument.
+        reference (object)
+            reference abundance pattern. Optional initialisation argument.
             Defaults to Asplund (2009) pattern.
```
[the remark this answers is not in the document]
stephenmwilkins: Fixed

src/synthesizer/abundances/abundance_patterns.py
stephenmwilkins: Fixed

stephenmwilkins:
> Do you want the user to be able to pass a custom elements class? If so the code currently doesn't allow this it only imports the elements dataclass.

Nope.

stephenmwilkins:
> A couple more comments from me:
> 
> It would be good to remove the init in the dataclasses and define them at the class level. These could then be overridden if a dictionary is passed.
> 
> i.e.
> 
> ```
> @dataclass
> class C:
>     a: int       # 'a' has no default value
>     b: int = 0   # assign a default value for 'b'
> ```
> 
> will automatically generate an init of the form:
> 
> ```
> def __init__(self, a: int, b: int = 0):
> ```

I've done this for elements but I'm not going to get round to it for the other classes anytime soon.

src/synthesizer/line.py
stephenmwilkins: This is not actually a dataclass, it's essentially a fancy way of holding some dictionaries, so doesn't need to be called. This would also be painful to turn into a dataclass since you wouldn't be able to do:

```
O3b = "O 3 4958.91A"
O3r = "O 3 5006.84A"
O3 = [O3b, O3r]
```

We could think of something better but I have no desire to do that at the moment.

WillJRoper: Hmm ok, probably not the place for any form of class then. More a module. Since this can't be changed by the user it'll be easy enough.

WillJRoper: Actually, it can be a dataclass. Have pushed.

WillJRoper:
Ok, things are looking ok. Really I'd like to clean things up a bit more but that'll have to be done in a future more focused PR. Only two things left:

- The TODO in the lines module doc string. 
- Chris's comment about the need for more description of the `Elements` class, i.e. what is it? What is it used for? Why should it never be changed? Any extra information that would give reader context for why it's done the way it is.

WillJRoper:
Address those 2 points and I will merge

stephenmwilkins:
Done those last changes I think.

WillJRoper (approved):

PR #547, opened 2024-03-19
Added an interactive grid explorer

stephenmwilkins:
Have added an interactive grid explorer to the `examples` directory. This allows a user to easily explore how the SED changes as a function of age and metallicity. The user can also interactively choose the spectra type and wavelength range.

## Issue Type

- [x] Enhancement

## Checklist


christopherlovell:
There are **a lot** of unrelated changes in this PR, in particular to abundances and lines. These should be isolated to a separate PR

christopherlovell (changes requested):
Unrelated changes should be implemented in a separate PR

WillJRoper:
I was going to say the same thing but I think this PR is just dependent on #546

WillJRoper:
Any particular reason this example is in a notebook?

WillJRoper (changes requested):
In principle, I really like this. Couple comments:

- I don't see why this isn't an example script. Are the widgets dependent on it being a notebook? 
- Can you use the package's `plot_spectra` function instead of reinventing the wheel? This can reduce your `plot_spectra` function here to a getter which just extracts a particular spectra for plotting. 
- This introduces an extra dependency solely for this one niche use. Either we drop it and close this PR or include instructions to install `ipywidgets` the first time this is run. I don't think (nor want) to add a blanket dependency for this.

stephenmwilkins:
> In principle, I really like this. A few comments:
> 
> * I don't see why this isn't an example script. Are the widgets dependent on it being a notebook?
> * Can you use the package's `plot_spectra` function instead of reinventing the wheel? This can reduce your `plot_spectra` function here to a getter which just extracts a particular spectra for plotting.
> * This introduces an extra dependency solely for this one niche use. Either we drop it and close this PR or include instructions to install `ipywidgets` the first time this is run. I don't think (nor want) to add a blanket dependency for this.

- I don't understand how the widgets would work outside of a notebook? 
- Yes, it could use `plot_spectra`.
- Don't add a blanket dependency and just let it fail if people don't have ipywidgets.

WillJRoper:
> - I don't understand how the widgets would work outside of a notebook?
> - Yes, it could use plot_spectra.
> - Don't add a blanket dependency and just let it fail if people don't have ipywidgets.

- Can they not work when run in standard ipython? Probably could do something more refined with a GUI engine but not worth it for now.
- "Yes, it ~~could~~ should use plot_spectra."
- I'll do something a bit more fancy than a flat failure with a try-except.

WillJRoper:
I've just implemented a safe exit when `ipywidgets` isn't found.

WillJRoper:
@stephenmwilkins, just a poke

stephenmwilkins:
> @stephenmwilkins, just a poke

I implemented a version using `plot_spectra` but it is much less smooth than the existing implementation. I'd prefer just to leave this version. My `plot_spectra` implementation exists but is commented out.

christopherlovell:
> > @stephenmwilkins, just a poke
> 
> I implemented a version using `plot_spectra` but it is much less smooth than the existing implementation. I'd prefer just to leave this version. My `plot_spectra` implementation exists but is commented out.

If it's just going to be commented out I advise we just get rid of it, we don't want commented out code not being run in the test suite and going out of date

WillJRoper:
It's no longer commented out. I'm not sure what smoother means but after passing the label and applying the figure text the `plot_spectra` version works fine. In fact, it works better since the other implementation cut of emission lines for both nebular and total due to the y axis limits which are derived automatically by the root plot_spectra function.

stephenmwilkins:
> It's no longer commented out. I'm not sure what smoother means but after passing the label and applying the figure text the `plot_spectra` version works fine. In fact, it works better since the other implementation cut of emission lines for both nebular and total due to the y axis limits which are derived automatically by the root plot_spectra function.

We don't want to set the y-axis limits automatically since it's useful to see how the normalisation of the spectra changes. Probably better is to allow the user to set y-axis limits with a slide, but TBH a hardcoded sensible range should be just fine.

By smoother, I meant the refresh. On my machine/VS code, there was a noticeable lag with `plot_spectra` compared to existing implementation.

WillJRoper:
I couldn't see a noticeable lag. Sure there was a slight delay. Having a slider for the y-axis sounds good. The previous hard coded limits only worked for `"incident"` with the other spectra types extending beyond the limits.

WillJRoper:
Can we merge this @christopherlovell?

christopherlovell (approved):

PR #552, opened 2024-03-26
Remove LineRatios, add line_ratios sub-module, and fix id bug 

stephenmwilkins:
- `LineRatios` class which functioned just as a container has been removed in favour of having a sub-module (`line_ratios`). 
- Also fixes an issue where the ID of hydrogen-alpha and beta were defined using vacuum wavelengths while other optical lines used air wavelengths (as standard).

## Issue Type

- [x] Bug
- [x] Enhancement

## Checklist


stephenmwilkins:
This is failing because the `test_grid` needs updating.

WillJRoper:
Changed the grid for this branch only. Once tested the download script needs editing and I'll fully swap the file on the server but since swapping that file will break everything else this can only happen once this branch is ready to merge.

WillJRoper:
@stephenmwilkins this now passes where it was failing before but is now failing in one of the notebooks for a different reason.

stephenmwilkins:
Ah... the AGN NLR and BLR grids also need updating. Seemingly cloudy17 (old grids) and cloudy23 (new grids) use different default wavelength meshes. We should probably address this separately using some kind of interpolating to bring two different spectra to the same wavelength grid when adding Sed objects but that's a task for another day. For now it would be good to also update the AGN grids. These are:

agnsed-limited_cloudy-c23.01-blr.hdf5
agnsed-limited_cloudy-c23.01-nlr.hdf5

In the Dropbox. @WillJRoper can you do this?

WillJRoper:
Sure, I'll make that change here.

WillJRoper:
AGN grids swapped out. Now failing in an example script:
```
Unexpected failing examples:

    ../../examples/grids/plot_lines.py failed leaving traceback:

    Traceback (most recent call last):
      File "/home/runner/work/synthesizer/synthesizer/examples/grids/plot_lines.py", line 39, in <module>
        line = grid.get_line(grid_point, "H 1 4862.69A")
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/synthesizer/grid.py", line 687, in get_line
        raise exceptions.InconsistentParameter(
    synthesizer.exceptions.InconsistentParameter: Provided line_id is not in the list of available lines. 

-------------------------------------------------------------------------------
```

stephenmwilkins:
Hopefully fixed now!

WillJRoper:
Ok so this looks good now. I can merge but doing so will break will mean anyone with old grids will no longer be able to run their scripts if they pull the latest main. I'm not sure how to approach this since we have a fair few people developing or using it now. 

@christopherlovell, @aswinpvijayan, @stephenmwilkins what should we do? Shall we put a warning out on Slack and deal with the fallout. In an ideal world we should go through and make everything backwards compatible but that could be a nightmare. This absolutely has to be the last breaking change without backwards compatibility and a deprecation warning if we merge as is.

stephenmwilkins:
Hold fire, I will need to replace the test_grid again. It won't break anything but it's another change that's needed.

stephenmwilkins:
Replacement grid sent to Will.

stephenmwilkins:
Apologies, I've made some additional improvements, mostly to the docs but I've also added some functions for converting between standard and vacuum wavelengths and done this automatically for `Line` objects.  Consequently I've had to add a new definition to units.

stephenmwilkins:
There is a new test_grid however the test_grid I originally gave and is currently implemented should (and appears to) work.

src/synthesizer/line.py
```diff
@@ -15,20 +15,133 @@
...
-from unyt import Angstrom
+from unyt import Angstrom, unyt_array
 
-from synthesizer import exceptions
+from synthesizer import exceptions, line_ratios
 from synthesizer.conversions import lnu_to_llam
 from synthesizer.units import Quantity
 
 
+def vacuum_to_air(wavelength):
```
christopherlovell, line 26: I think these conversion methods should be in the `conversions.py` top level module

src/synthesizer/line_ratios.py
```diff
@@ -0,0 +1,116 @@
+"""
+A module holding useful line ratios (e.g. R23) and diagrams (pairs of ratios),
+e.g. BPT-NII.
+
+Line ids and specifically the wavelength part here are defined using the cloudy
+standard, i.e. using vacuum wavelengths at <200nm and air wavelengths at
+>200nm.
+
+There is a parallel module that vacuum
```
christopherlovell, line 9: incomplete docs

src/synthesizer/line_ratios.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+
+# shorthand for common lines
+
+# standard names
+Ha = "H 1 6562.80A"
```
christopherlovell, line 15: I'm really confused as to why this is now a sub module with global attributes. I much preferred the previous implementation of a dedicated class. Alternatively, all of this information should just be defined on a line collection.

christopherlovell (changes requested):
I am very confused by the new `line_ratios` sub module, and the motivation for separating this out. All functionality for dealing with ratios between lines should be handled by a `LineCollection` object, and so all of this functionality could (and should?) be handled on that class. By separating this out it becomes unclear how this functionality is related, makes it more open to inconsistencies, and impossible for users to add their own ratios. The use of global variables is also bad practice

src/synthesizer/line.py
WillJRoper: +1

WillJRoper:
I've updated the grid now. All that needs doing is addressing @christopherlovell's review. I can see the argument for just attaching these to a `LineCollection` as class attributes (and the functions as methods). I do still prefer the submodule over the container class though.

src/synthesizer/line_ratios.py
christopherlovell: Fair. I preferred the previous setup as users could then add their own ratios, and then make the most of all the downstream calculating and plotting functionality, whereas the new format is fixed

WillJRoper: I'm not sure it could be modified previously, the class had no initialiser so there's no interface to get an instance of it and then modify it.

WillJRoper: It was just a namespace, exactly the same as it is now just without the class keyword.

src/synthesizer/line.py
stephenmwilkins: Done

src/synthesizer/line_ratios.py
stephenmwilkins: Done

src/synthesizer/line_ratios.py
stephenmwilkins: Indeed, there isn't much difference between the previous implementations.

If we really wanted people to be able to add their own line ratios etc. we could add the line_ratio module as an argument to LineCollection. HOWEVER, methods already exist on LineCollection to quickly measure arbitrary ratios or diagrams by providing lists of lines. I've just improved this to make it easier so now you can do:

`lines.get_ratio('R23')` or `lines.get_ratio([line1, line2])`

WillJRoper: That's a nice solution. My impression was that the line ratios were a static thing?

I'm pretty sure (but haven't run a test) that if we wanted the user to be able to add to the line ratios in either the old container or the submodule structures we'd end up in one of those situations where we always get a copy and modify the original every time. That could lead to some really opaque bugs down the line. Better to have the line collection methods for arbitrary things.

WillJRoper:
New grid causes docs error:
```
---------------------------------------------------------------------------
UnrecognisedOption                        Traceback (most recent call last)
Cell In[15], line 8
      6 grid_point = (ia, iZ)
      7 lines = grid.get_lines(grid_point)
----> 8 x_, y_ = lines.get_diagram(diagram_id)
      9 x.append(x_)
     10 y.append(y_)

File /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/synthesizer/line.py:453, in LineCollection.get_diagram(self, diagram_id)
    451     # check if ratio_id exists
    452     elif diagram_id not in self.available_ratios:
--> 453         raise exceptions.UnrecognisedOption(
    454             """LineCollection is missing the lines required for
    455             this diagram"""
    456         )
    458     ab, cd = self.line_ratios.diagrams[diagram_id]
    460 # otherwise interpret as a list

UnrecognisedOption: LineCollection is missing the lines required for
                    this diagram 
```

stephenmwilkins:
> New grid causes docs error:
> 
> ```
> ---------------------------------------------------------------------------
> UnrecognisedOption                        Traceback (most recent call last)
> Cell In[15], line 8
>       6 grid_point = (ia, iZ)
>       7 lines = grid.get_lines(grid_point)
> ----> 8 x_, y_ = lines.get_diagram(diagram_id)
>       9 x.append(x_)
>      10 y.append(y_)
> 
> File /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/synthesizer/line.py:453, in LineCollection.get_diagram(self, diagram_id)
>     451     # check if ratio_id exists
>     452     elif diagram_id not in self.available_ratios:
> --> 453         raise exceptions.UnrecognisedOption(
>     454             """LineCollection is missing the lines required for
>     455             this diagram"""
>     456         )
>     458     ab, cd = self.line_ratios.diagrams[diagram_id]
>     460 # otherwise interpret as a list
> 
> UnrecognisedOption: LineCollection is missing the lines required for
>                     this diagram 
> ```

I think I introduced a bug; I think it's fixed now.

christopherlovell (approved):

PR #569, opened 2024-04-05
Agn improvements

### LINKED ISSUE #567: Masking issue in blackhole modelling
The current use of masks means that if the disc is behind the torus the `disc_incident` emission is set to zero. This problematic since then the torus emission is zero. This can be fixed by setting the mask for the `disc_incident` creation to `None` and multiplying `self.spectra["disc_incident"]` by `mask` when calculating the total transmitted spectra. 

There was also a bug where a zero spectra was set as `np.zeros(grid.lam)` instead of `np.zeros(len(grid.lam))`

Fix incoming.

stephenmwilkins:
Fixes #567 

## Issue Type

- [x] Bug
- Document
- Enhancement


WillJRoper (changes requested):
This can't be merged since it includes code from other unmerged branches. This should be branched off main or made draft until it can be merged.

stephenmwilkins:
Argh... I don't understand what's happened here. I thought I made a branch from `main`. I'll fix later.

stephenmwilkins:
Fixing in another branch

PR #570, opened 2024-04-05
fixed issues with masks

stephenmwilkins:
Fixing a bug were high-inclinations (i.e. obscured by the torus) didn't work properly.

- zeroed spectra were initialised with `np.zeros(grid.lam)` instead of `np.zeros(len(grid.lam))`
- Torus modelling needs the isotropic disc_incident spectra so have added this and used it.
- We now saved both the masked and unmasked `disc_incident` spectra. The masked is used to generate the "disc_escaped" spectra. Since `disc_incident` and `disc_incident_masked` are equivalent at low-inclination and `disc_incident_masked` is zero otherwise there is probably a better of way of handling this, for example, by multiplying `self.spectra["disc_incident"]` by the mask when used. However, this doesn't seem to work as the shapes are incorrect which is slightly puzzling and may point to a deeper issue.

## Issue Type

- Bug
- Document
- [x] Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/components/blackhole.py
+++ b/src/synthesizer/components/blackhole.py
@@ -446,7 +446,26 @@ class BlackholesComponent:
         # Reset the cosine_inclination to the original value.
         self.cosine_inclination = prev_cosine_inclincation
 
+        # This is the true incident disc emission, i.e. not including the
+        # mask.
         self.spectra["disc_incident"] = Sed(
+            lam,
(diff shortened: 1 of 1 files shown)


stephenmwilkins:
Not yet fixed because there is another bug I'm trying to track down.

WillJRoper (approved):

PR #573, opened 2024-04-08
Rename blackhole methods

### LINKED ISSUE #568: Naming of spectra generation methods in `components/blackhole` needs to relate to the model
Naming of spectra generation methods in `components/blackhole` needs to relate to the model. For example, `_get_spectra_torus` only relates to the `UnifiedAGN` model. Unless we store these methods elsewhere (my preference) they need to have clear names, e.g. `_get_spectra_unifiedagn_torus`. 

Similarly, `get_spectra_incident` needs a check on which model to use beyond excluding just `Template`.

stephenmwilkins:
If we are really intent to keep all the spectra generation methods for different blackhole models in `component/blackholes.py` we need to name them sensibly since e.g. the torus modelling in `UnifiedAGN` may differ from the torus modelling in a later model. This PR renames the methods used by the `UnifiedAGN` model, adds an if statement for that model, and raises an exception if the emission_model is not implemented (i.e. not Template or UnifiedAGN").

- Draft until #570 merged.
- Closes #568.

## Issue Type

- Bug
- Document
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/components/blackhole.py
+++ b/src/synthesizer/components/blackhole.py
@@ -446,7 +446,26 @@ class BlackholesComponent:
         # Reset the cosine_inclination to the original value.
         self.cosine_inclination = prev_cosine_inclincation
 
+        # This is the true incident disc emission, i.e. not including the
+        # mask.
         self.spectra["disc_incident"] = Sed(
+            lam,
+            self.generate_lnu(
+                emission_model,
+                emission_model.grid["nlr"],
+                spectra_name="incident",
+                line_region="nlr",
+                fesc=0.0,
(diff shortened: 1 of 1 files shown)


WillJRoper:
This is exactly what I want to avoid. We need to talk about this offline. We don't have "other models" right now. When we do we can abstract the methods on the component. For now we have 2 methods which work in this framework. I have thought about this and have a suggestion for how we should approach this but we need to talk before we do anything.

I'm closing this PR. Can you shift to some other open stuff instead of sweeping changes to the AGN until we have discussed this in detail.

Issue #534, opened 2024-02-23
Optionally add CMB heating to dust emission 


stephenmwilkins:
Add the effect of CMB heating on thermal dust emission. See e.g. [https://ui.adsabs.harvard.edu/abs/2013ApJ...766...13D/abstract](https://ui.adsabs.harvard.edu/abs/2013ApJ...766...13D/abstract)

PR #574, opened 2024-04-08
Option to add enchancement of IR luminosity due to CMB heating

aswinpvijayan:
## Issue Type
- Enhancement

Using the prescription in da Cunha+2013, added the option to boost the IR luminosity from heating by CMB.

This closes issue #534

## Checklist


WillJRoper (changes requested):
This looks good to me but I'm not an expert on it. Two things:

1. Don't bother now as we can do it on the typing PR but avoid `# type: ignore` at all costs.
2. Can you put some comments surrounding the cmb heating/temperature calculations just explaining the rational behind the approach and what each step is doing.

src/synthesizer/dust/emission.py
```diff
@@ -100,21 +123,54 @@ class Blackbody(EmissionBase):
...
+
+            z (float)
+                Redshift of the galaxy
+
         """
+        # emmissivity of true blackbody is 1
+        emissivity = 1.0
+        _temperature: unyt_quantity
+        cmb_factor: float
+        if cmb_heating:
```
WillJRoper, line 155: Like here, what is this whole block doing, what is going on with the derivation in this block. Some more context would help.

WillJRoper: Also, correct me if I'm wrong but isn't this the same for all models? Could we not make a private method for this and attach it to the parent class to reduce boilerplate?

WillJRoper:
@aswinpvijayan can you take a look at this so we can get it in?

WillJRoper (approved):

PR #575, opened 2024-04-08
Consolidate blackholes

stephenmwilkins:
Consolidated blackholes sed generation. All blackholes are now expressed as arrays allowing us to use one set of methods to generate spectra. 

- killed existing sed generation methods on `component/blackholes`, including `generate_lnu`.
- when a parametric blackhole is initialised the arguments are converted to a blackhole array with shape (1,).
- moved particle sed methods (e.g. `get_particle_spectra_intrinsic`) to `component/blackholes` so they can be accessed by parametric blackholes. This is slightly icky since these methods with the name particle are no longer in `particle/blackholes`.
- added new methods `get_spectra_intrinsic` and `get_spectra_attenuated` which simply call e.g. `get_particle_spectra_intrinsic` and sum the spectra.
- fixed a bug wherein dust attenuation didn't work for particle blackholes.

## Issue Type

- [x] Bug
- Document
- [x] Enhancement

## Checklist


stephenmwilkins:
I appreciate this is a bit nuclear but I think it is an improvement in that we now only need to maintain one set of methods.

src/synthesizer/parametric/blackholes.py
```diff
@@ -78,6 +78,21 @@ def __init__(
                 here to override the defaults of the emission models.
         """
 
+        if bolometric_luminosity is not None:
```
WillJRoper, line 81: `utils.value_to_array` handles this block in one liners

WillJRoper (changes requested):
I see what you're getting at with this but this isn't the solution. You've already mentioned the ickyness of the particle naming scheme but that's just the tip of the iceberg. We still need to maintain having particle *and* non-particle methods because one is integrated and one is per particle (they have a specific meaning on the front end). The backend can deal with only "particles" of course, but I don't think that requires the naming convention to always use particles. In fact, I'd be inclined to have `_generate_lnu` maintain it's existence for integrated spectra but all it needs to do is massage the inputs into arrays and call the true particle version.

This also doesn't address the root issue we discussed. The current form of `generate_lnu` (particle or otherwise) assumes a certain set of inputs. These methods based on our discussion should be offloaded to the emission model but some clever structuring of that process is required to ensure adding new models is easy enough without meaning the developer needs to write every single one of these methods (especially `generate_lnu` which does a lot of specialist stuff for the C extension).

WillJRoper:
I really think we need a separate focused discussion about the approach to this next week.

WillJRoper:
From Slack (Will):

- We have the front-facing methods on the component/particle for each type of spectra (like everywhere else:
    - get_spectra_total
    - get_spectra_torus
    - get_spectra_disc
    - get_spectra_blr
    - get_spectra_nlr
    - get_spectra_intrinsic
    - etc. + particle versions
- The emission models should define the list of private methods they define.
- The component/particle then checks if the corresponding method being called exists on the emission model. If so it calls the private version if not it creates a zero array and warns the user (if appropriate).
- The emission model should be nothing more than a Grid analogue for encapsulating the more complex modelling. It should not contain any front-facing methods for spectra generation, this is handled by components in Synthesizer.
- The emission model docs should be heavily modified to only describe their definition, extension and how it works like a Grid.

PR #585, opened 2024-04-15
Fix dust curve kwargs in CF00

christopherlovell:
Main fix is to remove the `dust_curve`  specified in the default arguments inside `get_spectra_CharlotFall`. Also tidies up some of the label string concatenation.

## Issue Type
- Bug

## Checklist

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -197,7 +197,7 @@ class StarsComponent:
         sed = Sed(grid.lam, lnu)
 
         # Store the Sed object
-        self.spectra[label + "incident"] = sed
+        self.spectra[f"{label}_incident"] = sed
 
         return sed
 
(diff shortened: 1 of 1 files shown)


src/synthesizer/components/stellar.py
```diff
@@ -197,7 +197,7 @@ def get_spectra_incident(
         sed = Sed(grid.lam, lnu)
 
         # Store the Sed object
-        self.spectra[label + "incident"] = sed
+        self.spectra[f"{label}_incident"] = sed
```
WillJRoper, line 200: I'm not sure these fixes are right? Doesn't `label` include a trailing underscore by design? Explicitly because you can also have a incident on it's own? Or is that not true for CFOO?

WillJRoper: Oh wait, I just the change below. Is this specifically because you always have young and old spectra in CF00?

christopherlovell: Actually, you're absolutely right. Commit below fixes this

Issue #377, opened 2023-11-06
Add method to LineCollection object to create Spectra


stephenmwilkins:
It will be useful to be able to generate a line contribution spectrum from a LineCollection instance. This would allow us e.g. to create more appropriate spectra when downsampling or converting the wavelength grid.

WillJRoper:
@stephenmwilkins can you give more details as to what you mean? It's not clear to me what you are asking for.

stephenmwilkins:
If you have the luminosities of a set of lines you should be able to create a spectrum just using those lines, assuming you know the wavelengths. To do this properly you need to assume some line width but in the default instance you can assume all the photons are in a single resolution element.

WillJRoper:
So that's effectively just a scatter plot?

stephenmwilkins:
You could represent it by a scatter plot (wavelength_i, luminosity_i) but what we want is an actual spectrum. Obviously (depending on the resolution) some resolution elements (bins) will have a contribution from more than one line.

WillJRoper:
Ah ok, I'm with you now

PR #587, opened 2024-04-16
Create sed from line collection

stephenmwilkins:
Adds a method to `LineCollection` to create an `Sed` from the collection of lines. 

- lines.ipynb doc notebook to include example.
- Removed example script since now deprecated.

41702955-3883-46f6-bdb6-73e5e32abde4

As you can see in the above figure (output from lines.ipynb) this doesn't quite work: the derived spectra is slightly higher than the linecont spectra provided by cloudy. Note there is very little difference for all the strong lines, however for some faint UV lines there is a clear difference. I don't believe this is a consequence of the method however, rather one of several possibilities:

- Currently we are using intrinsic line luminosities, but these should probably be swapped to emergent line luminosities which include an additional source of dust. I'm trying this now but I'm puzzled because I think the consequence will overshoot this affect.
- cloudy line wavelengths are not quite correct.

## Issue Type

- [x] Enhancement

## Checklist


stephenmwilkins:
As I somewhat expected switching to use emergent line luminosities doesn't fix the issue. The spectra inferred from the line collection now sits a little below that reported directly by cloudy, at least for most of the strong lines.

Unless anyone spots an obvious bug I think this is simply some quirk of `cloudy`.


845c6187-fe8a-4e2d-8187-15c6f3911a98

WillJRoper:
Where are we with this @stephenmwilkins? Are you saying we should merge it with the issue and chalk it up to `cloudy` quirks (I'm not keen in including something broken), or should we just close it until we can get them to match?

WillJRoper:
@stephenmwilkins Poke

WillJRoper:
Closing this because it's fallen too far behind and is going to be a real pig to sort out the conflicts. The issue still exists (#377) so, we'll raise a new PR for it.

Note also that the actual sorting of lines onto an `Sed` should have used interpolation and should do on the future branch.

PR #593, opened 2024-04-19
Ensure filter resampling on new filter collection

christopherlovell:
DWISOTT. Was not called by default on instantiation when it should have been.

## Issue Type
- Bug

## Checklist


WillJRoper:
Sorry, I moved the resample. It had to be done after all the include methods but could also feasibly be applied to a loaded set of filters if `new_lam` has been passed.

WillJRoper (approved):

PR #599, opened 2024-05-03
Implement particle based line emission and misc fixes

### LINKED ISSUE #538: Lines for particle based galaxies do not work
The methods for calculating line quantities for particle based galaxies do not currently work. 

This needs a c extension analogous to the SED for getting the grid weights for each particle.

WillJRoper:
This PR implements the machinery to get lines from particle-based stars.

- Both integrated and per particle C extensions now exist. These are called from the `get_lines*` and `get_particle_lines*` methods. 
- `LineCollection`s can now be concatenated.
- line ids are now always strings instead of strings or lists. A list can be passed but a deprecation warning is printed.

This also includes some fixes:
- Allowing both strings and lists/tuples for line_ids made the code a mess and bug-prone. Now, internally only strings are handled for >=doublets. A deprecation warning is printed if a list or tuple is passed before conversion.
- If a user called an intrinsic line method and later called a screen or attenuated method with a different set of lines the code would fail since it only checked if the `"intrinsic"` key existed and not if the requested lines already existed.
- Blanket calling `mean` and `sum` on the inputs to a line breaks for arrays of luminosity, continuum and wavelengths (i.e. in the per particle case). This required some rewriting to ensure the input dimensions are respected.
- `LineCollections` attached to a component are now correctly updated instead of being overwritten.
- `Line.__str__` fixed to play nice with arrays.

Some further clean-up is needed in the initialisation of lines and line collections but that will come in the future.

Closes #538.

## Issue Type

- Enhancement

## Checklist


stephenmwilkins:
Looks good, main thought (as mentioned on Slack) is that we need a working notebook example.

WillJRoper:
> Looks good, main thought (as mentioned on Slack) is that we need a working notebook example.

A notebook specific to lines will come. Possibly better handled by a non-notebook example since it's a bit more specific but we can see what the docs look like.

src/synthesizer/line.py
```diff
@@ -364,34 +403,37 @@ def __next__(self):
...
-            a = [a]
-        if isinstance(b, str):
-            b = [b]
-
-        return np.sum([self.lines[_line].luminosity for _line in a]) / np.sum(
-            [self.lines[_line].luminosity for _line in b]
-        )
+        return np.sum(
+            [self.lines[_line].luminosity for _line in line1]
+        ) / np.sum([self.lines[_line].luminosity for _line in line2])
```
stephenmwilkins, line 432: At the moment this sums over all axes right? So if we have an array (i.e. for different particles) they are automatically summed. I think we want this to be include axis=1 (or is it axis=0).

WillJRoper: Good catch! Fixed now.

stephenmwilkins (approved):

PR #601, opened 2024-05-06
More intuitive multi-line `Line` objects (e.g. doublets).

WillJRoper:
A pretty minor improvement overall but this PR improves the intuitiveness of multi-line Lines. Previously a list or array of values for the individual lines making up a composite line was passed and stored. This required some logical gymnastics to make sure everything ended up where it needed to be in different situations. 

- Now a `Line` can be initialised using either an arbitrary number of `Line`s or the id, wavelength, luminosity and continuum that defines an individual line.
- Any individual lines used to create a new `Line` are stored in a `indiviudal_lines` attribute.
- A user-facing method to combine lines has been included for adding >2 lines together.
- `Line` addition has been fixed. It would have previously "run" but never would have given the correct answer unless you were adding 2 lines with the exact same ID... making it somewhat redundant.
- The `Line` definition is now fully documented.

Dependant on #599.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -1204,7 +1204,7 @@ class StarsComponent:
         elif isinstance(line_ids, (list, tuple)):
             # Convert all tuple or list line_ids to strings
             line_ids = [
-                ",".join(line_id)
+                ", ".join(line_id)
                 if isinstance(line_id, (list, tuple))
                 else line_id
                 for line_id in line_ids
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -1344,15 +1344,15 @@ class StarsComponent:
             )
 
             # Apply attenuation
-            luminosity = intrinsic_line._luminosity * T_nebular * T_stellar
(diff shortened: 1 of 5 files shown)


WillJRoper:
Still dependant on #599 but since, that ones seems near merging I'm also opening this for review.

stephenmwilkins:
This is slightly out of the scope of the PR but might make sense to implement now. I think we may find a need to be able to initialise lines without continuum luminosities. Basically, not providing a continuum shouldn't make the code fall over and should set `equivalent_width` to `None`.

WillJRoper:
> This is slightly out of the scope of the PR but might make sense to implement now. I think we may find a need to be able to initialise lines without continuum luminosities. Basically, not providing a continuum shouldn't make the code fall over and should set `equivalent_width` to `None`.

What's the use case for that? That'll break a lot of the C stuff I just added. 

Definitely out of scope though so raise an issue. I'm working on slowly cleaning up the Lines stuff in very modular PRs, this is only the first, so I can make that change if it is useful.

PR #604, opened 2024-05-09
Particle line fix

stephenmwilkins:
Fixes a pair of bugs in the `get_particle_line_attenuated` and `get_particle_line_screen`.

- The attenuation felt by the line emission should *probably* just be `T_nebular` not `T_nebular * T_stellar`. I can see an argument for `T_nebular * T_stellar` if there is attenuation between the source and the nebular but that isn't consistently modelled by cloudy.
- Probably related to the above. For a screen `tau_v_nebular = tau_v_screen = tau_v`.
- A bigger bug was caused by the fact that `T_nebular` and `T_stellar` have shapes `(npart, 1)` for a line, not `(npart)` as assumed by the calculation.
- Also improved `component.stars.get_line_attenuated` by raising an exception if `tau_v_*` is not a float (e.g. if the user tries to provide an array as I did.

## Issue Type

- [x] Bug

## Checklist

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -1273,7 +1273,7 @@ class StarsComponent:
             fesc (float)
                 The Lyman continuum escaped fraction, the fraction of
                 ionising photons that entirely escaped.
-            tau_v_BS (float)
+            tau_v_nebular (float)
                 V-band optical depth of the nebular emission.
             tau_v_stellar (float)
                 V-band optical depth of the stellar emission.
@@ -1323,6 +1323,18 @@ class StarsComponent:
                 method=method,
             )
 
+        # Check that tau_v_nebular and tau_v_stellar are floats and raise
+        # an exception otherwise.
+        if not isinstance(tau_v_nebular, float):
+            raise exceptions.InconsistentArguments(
+                """ tau_v_* must be a float (i.e. single value)."""
+            )
+
+        if not isinstance(tau_v_stellar, float):
+            raise exceptions.InconsistentArguments(
+                """ tau_v_* must be a float (i.e. single value)."""
+            )
+
         # Get the intrinsic lines now we're sure they are there
         intrinsic_lines = self.lines["intrinsic"]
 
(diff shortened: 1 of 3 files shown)


stephenmwilkins:
Actually, right now this breaks if you provide a single value of `tau_v` instead of an array.

stephenmwilkins:
This still doesn't work properly for the reason explained in #607

src/synthesizer/components/stellar.py
```diff
@@ -1323,6 +1323,18 @@ def get_line_attenuated(
                 method=method,
             )
 
+        # Check that tau_v_nebular and tau_v_stellar are floats and raise
+        # an exception otherwise.
+        if not isinstance(tau_v_nebular, float):
+            raise exceptions.InconsistentArguments(
+                """ tau_v_* must be a float (i.e. single value)."""
```
WillJRoper, line 1330: Not doc strings.

src/synthesizer/components/stellar.py
```diff
@@ -1323,6 +1323,18 @@ def get_line_attenuated(
...
+            )
+
+        if not isinstance(tau_v_stellar, float):
+            raise exceptions.InconsistentArguments(
+                """ tau_v_* must be a float (i.e. single value)."""
```
WillJRoper, line 1335: same here

src/synthesizer/particle/stars.py
```diff
@@ -1788,8 +1788,14 @@ def get_particle_line_attenuated(
                 tau_v_stellar, intrinsic_line._wavelength
             )
 
+            if not np.isscalar(T_nebular):
```
WillJRoper, line 1791: I don't think the logic is needed here. The transmission is guaranteed to be returned in the strange shape that was causing issues. 

Porbably better to have `get_transmission` to take some sort of shape argument to handle this situation rather than fix it outside that method. That way we won't have to play this game anywhere else we have this problem.

WillJRoper (changes requested):

src/synthesizer/components/stellar.py
stephenmwilkins: Fixed.

src/synthesizer/components/stellar.py
stephenmwilkins: Fixed.

src/synthesizer/particle/stars.py
WillJRoper: It's probably a simple case of modifying `get_transmission` to remove the wavelength axis when a single wavelength is passed.

stephenmwilkins: OK, I agree. I've fixed `get_transmission` instead to return the correct shape depending on whether tau_v and lam are arrays.

WillJRoper (approved):

PR #611, opened 2024-05-13
added method on stars to generate attenuated spectra and produce summed Lines and LineCollections

### LINKED ISSUE #610: Add a sum method on Line and LineCollection
In analogy with `Sed` add a sum method on `Line` and `LineCollection` to return integrated quantities.

stephenmwilkins:
Adds a new method on `particles.stars` to generate dust attenuated spectra, in this case for a simple screen.

I have mistakenly also implemented line summing on this branch (fulfilling #610) too. No point separating it.

Closes #610.

## Issue Type

- [x] Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/line.py
+++ b/src/synthesizer/line.py
@@ -403,6 +403,18 @@ class LineCollection:
             # Return the filter
             return self.lines[self.line_ids[self._current_ind - 1]]
 
+    def sum(self):
+        """
+        For collections containing lines from multiple particles calculate the
+        integrated line properties and create a new LineCollection object.
+        """
+
+        summed_lines = {}
+        for line_id, line in self.lines.items():
+            summed_lines[line_id] = line.sum()
+
+        return LineCollection(summed_lines)
+
     def _get_ratio(self, line1, line2):
         """
         Measure (and return) a line ratio.
@@ -620,6 +632,9 @@ class Line:
                 " passed, or an arbitrary number of Lines to combine"
             )
 
+        # save line_id
+        self.line_id = line_id
+
         # Initialise an attribute to hold any individual lines used to make
         # this one.
         self.individual_lines = lines if len(lines) > 0 else [self]
@@ -761,6 +776,19 @@ class Line:
                 New instance of Line containing both lines.
         """
         return Line(self, second_line)
+
+    def sum(self):
+        """
+        For objects containing lines of multiple particles sum them to produce
+        the integrated quantities.
+        """
+
+        return Line(
+            line_id=self.line_id,
+            wavelength=self.wavelength,
+            luminosity=np.sum(self.luminosity),
+            continuum=np.sum(self.continuum),
+        )
 
     def get_flux(self, cosmo, z):
         """
--- a/src/synthesizer/line.py
+++ b/src/synthesizer/line.py
@@ -632,9 +632,6 @@ class Line:
                 " passed, or an arbitrary number of Lines to combine"
             )
 
-        # save line_id
-        self.line_id = line_id
-
         # Initialise an attribute to hold any individual lines used to make
         # this one.
         self.individual_lines = lines if len(lines) > 0 else [self]
@@ -784,7 +781,7 @@ class Line:
         """
 
         return Line(
-            line_id=self.line_id,
+            line_id=self.id,
             wavelength=self.wavelength,
             luminosity=np.sum(self.luminosity),
             continuum=np.sum(self.continuum),
--- a/src/synthesizer/particle/stars.py
+++ b/src/synthesizer/particle/stars.py
@@ -1775,6 +1775,7 @@ class Stars(Particles, StarsComponent):
         dust_curve_stellar=PowerLaw(slope=-1.0),
         mask=None,
         method="cic",
+        label="",
     ):
         """
         Get a LineCollection containing attenuated lines for each particle.
@@ -1814,6 +1815,11 @@ class Stars(Particles, StarsComponent):
             LineCollection
                 A dictionary like object containing line objects.
         """
+
+        # Make a dummy mask if none has been passed
+        if mask is None:
+            mask = np.ones(self.nparticles, dtype=bool)
+
         # If the intrinsic lines haven't already been calculated and saved
         # then generate them
         if "intrinsic" not in self.particle_lines:
--- a/src/synthesizer/particle/stars.py
+++ b/src/synthesizer/particle/stars.py
@@ -1605,6 +1605,80 @@ class Stars(Particles, StarsComponent):
 
         return reprocessed
 
+    def get_particle_spectra_screen(
+        self,
+        grid=None,
+        fesc=0.0,
+        tau_v=None,
+        dust_curve=PowerLaw(slope=-1.0),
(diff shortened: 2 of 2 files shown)


src/synthesizer/line.py
```diff
@@ -620,6 +632,9 @@ def __init__(
                 " passed, or an arbitrary number of Lines to combine"
             )
 
+        # save line_id
+        self.line_id = line_id
```
WillJRoper, line 636: Why have you introduced this? `self.id` contains the line ID and is set in the private constructor methods.

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,80 @@ def get_particle_spectra_reprocessed(
 
         return reprocessed
 
+    def get_particle_spectra_screen(
```
WillJRoper, line 1608: Could we not just use a generic attenuated method and include a `label` argument? Otherwise, we will have to derive a bunch of different spectra methods like the integrated version which encompasses all possible masks etc. 

This is why its been left on `Sed`s up until now, combining different masks and spectra is really non-trivial without introducing lots of code.

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,80 @@ def get_particle_spectra_reprocessed(
...
+
+        Returns:
+            Sed
+                An Sed object containing the attenuated spectra.
+        """
+
+        # If the reprocessed spectra haven't already been calculated and saved
+        # then generate them.
+        if "intrinsic" not in self.particle_spectra:
+            self.get_particle_spectra_reprocessed(
```
WillJRoper, line 1651: This will fail if the intrinsic spectra was made with a different mask to the current call (another reason for the label).

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,80 @@ def get_particle_spectra_reprocessed(
...
+                grid,
+                fesc=fesc,
+                mask=mask,
+                method=method,
+            )
+
+        # If tau_v is None use the tau_v on stars otherwise raise exception.
+        if not tau_v:
```
WillJRoper, line 1659: Please be more explicit: `if tau_v is not None`

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,80 @@ def get_particle_spectra_reprocessed(
...
+            if hasattr(self, "tau_v"):
+                tau_v = self.tau_v
```
WillJRoper, line 1661: This will need the mask applied to it.

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,80 @@ def get_particle_spectra_reprocessed(
...
+                    """
+                    tau_v must either be provided or exist on stars for
+                    attenuated spectra to be calculated.
+                    """
+                )
+
+        intrinsic_spectra = self.particle_spectra["intrinsic"]
+
+        # apply attenuated and save Sed object
+        self.particle_spectra["attenuated"] = (
```
WillJRoper, line 1673: Can apply the label here. You can see this approach in some of the integrated methods on a component.

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,80 @@ def get_particle_spectra_reprocessed(
...
+                method=method,
+            )
+
+        # If tau_v is None use the tau_v on stars otherwise raise exception.
+        if not tau_v:
+            if hasattr(self, "tau_v"):
+                tau_v = self.tau_v
+            else:
+                raise exceptions.InconsistentArguments(
+                    """
```
WillJRoper, line 1664: This should be normal strings broken over lines if needed. Doc strings aren't for long strings they are for documentation. This will print weirdly.

WillJRoper (changes requested):

src/synthesizer/line.py
stephenmwilkins: Argh... didn't see that. Removed.

src/synthesizer/particle/stars.py
stephenmwilkins: Yes, added label.

src/synthesizer/particle/stars.py
stephenmwilkins: Fixed.

src/synthesizer/particle/stars.py
stephenmwilkins: Fixed.

src/synthesizer/particle/stars.py
stephenmwilkins: Done.

src/synthesizer/particle/stars.py
stephenmwilkins: Done.

src/synthesizer/particle/stars.py
stephenmwilkins: Done.

src/synthesizer/particle/stars.py
WillJRoper: Cool. One remaining concern is how best to provide an interface to combine different attenuated spectra (i.e. old and young) generated via this.

src/synthesizer/particle/stars.py
```diff
@@ -1605,6 +1605,81 @@ def get_particle_spectra_reprocessed(
...
+            attenuated
+
+            if fesc>0:
+                escaped
+
+        Returns:
+            Sed
+                An Sed object containing the attenuated spectra.
+        """
+
```
WillJRoper, line 1648: Need an 
```
if label[-1] != "_":
    label = f"{label}_"
```

here.

WillJRoper: Also need to create a True everywhere mask `if mask is None`

WillJRoper (changes requested):
Last couple bits.

WillJRoper:
I'm resolving the comment but we need to think about how to combine different spectra together when not working at the `Sed` level. It may be the easiest way of addressing that is using a method specifically for having young and old contributions and just recommending anything more complex is done at the `Sed` level.

src/synthesizer/particle/stars.py
stephenmwilkins: Hmm... do we really want the label thing? Won't that fail is label=""?

WillJRoper: Ah my bad:
```
if len(label) > 0 and label[-1] != "_":
    label = f"{label}_"
```

stephenmwilkins: I appreciate it's nice to enforce labels include an underscore, but don't we need to add this all over the place.

WillJRoper: It's already handled in the integrated stuff that takes a label

stephenmwilkins: Won't it need adding in every method that takes a label?

stephenmwilkins: OK, done I think

WillJRoper (approved):

PR #613, opened 2024-05-14
Fix some issues with attenuated spectra

stephenmwilkins:
`get_particle_spectra_screen()` was bugged previously. Because this method sets `tau_v` with a mask, we need to define a dummy (empty) mask if one is not provided. Normally this happens elsewhere.

## Issue Type

- [x] Bug
- [ ] Document
- [ ] Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/dust/attenuation.py
+++ b/src/synthesizer/dust/attenuation.py
@@ -146,6 +146,7 @@ class AttenuationLaw:
         # Include the V band optical depth in the exponent
         # For a scalar we can just multiply but for an array we need to
         # broadcast
+
         if np.isscalar(tau_v):
             exponent = tau_v * tau_x_v
         else:
--- a/src/synthesizer/particle/stars.py
+++ b/src/synthesizer/particle/stars.py
@@ -798,6 +798,7 @@ class Stars(Particles, StarsComponent):
         grid,
         spectra_name,
         fesc=0.0,
+        mask=None,
         young=None,
         old=None,
         verbose=False,
@@ -914,7 +915,8 @@ class Stars(Particles, StarsComponent):
                 )
 
         # Get particle age masks
-        mask = self._get_masks(young, old)
+        if mask is None:
+            mask = self._get_masks(young, old)
 
         # Ensure and warn that the masking hasn't removed everything
         if np.sum(mask) == 0:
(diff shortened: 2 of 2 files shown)


src/synthesizer/dust/attenuation.py
```diff
@@ -146,6 +146,7 @@ def get_transmission(self, tau_v, lam):
         # Include the V band optical depth in the exponent
         # For a scalar we can just multiply but for an array we need to
         # broadcast
+
```
WillJRoper, line 149: Why the added newline?

src/synthesizer/particle/stars.py
```diff
@@ -798,6 +798,7 @@ def generate_particle_lnu(
         grid,
         spectra_name,
         fesc=0.0,
+        mask=None,
```
WillJRoper, line 801: We need to commit to one or the other I think. Adding both is unnecessarily introducing complexity.

However, which is preferable I am not convinced of.

src/synthesizer/particle/stars.py
```diff
@@ -952,7 +954,7 @@ def generate_particle_line(
         line_id,
         fesc,
         mask=None,
-        method="cic",
+        grid_assignment_method="cic",
```
WillJRoper, line 957: Good catch.

src/synthesizer/particle/stars.py
```diff
@@ -1616,7 +1618,6 @@ def get_particle_spectra_screen(
         tau_v=None,
         dust_curve=PowerLaw(slope=-1.0),
         mask=None,
-        method="cic",
```
WillJRoper, line 1619: Doesn't this need replacing with `grid_assignment_method`

src/synthesizer/particle/stars.py
```diff
@@ -1662,7 +1667,6 @@ def get_particle_spectra_screen(
                 grid,
                 fesc=fesc,
                 mask=mask,
-                method=method,
```
WillJRoper, line 1665: Same here.

WillJRoper (changes requested):
Obviously, some discussion to be had about what way we should be doing this. I'm not convinced one way or another.

src/synthesizer/particle/stars.py
WillJRoper: That being said. We should not remove the old way right now and should raise a deprecation warning.

stephenmwilkins:
Superseded

PR #615, opened 2024-05-14
Fix masking for particle_spectra_screen method

stephenmwilkins:
Fixes `get_particle_spectra_screen` method by including a call to `_get_mask`.

## Issue Type

- [x] Bug

## Checklist

### DIFF
--- a/src/synthesizer/particle/stars.py
+++ b/src/synthesizer/particle/stars.py
@@ -1615,8 +1615,8 @@ class Stars(Particles, StarsComponent):
         fesc=0.0,
         tau_v=None,
         dust_curve=PowerLaw(slope=-1.0),
-        mask=None,
-        method="cic",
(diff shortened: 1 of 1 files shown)


WillJRoper:
We are no longer going to introduce `young` and `old` in the lines. Only maintain them for a period where they already exist.

stephenmwilkins:
> We are no longer going to introduce `young` and `old` in the lines. Only maintain them for a period where they already exist.

That's fine. This is just bringing `get_particle_spectra_screen` in-line with the other spectra generation methods and allowing it to actually work. That is, it doesn't concern lines.

I suggest merging this so we have something that works and is consistent.

WillJRoper (approved):
My bad. I couldn't see it was spectra in the diff and assumed it was lines.

PR #618, opened 2024-05-14
Add a method to calculate integrated spectra for all the available particle_spectra

stephenmwilkins:
Adds a method to calculate integrated spectra for all the available particle_spectra.

## Issue Type

- Bug
- Document
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/particle/stars.py
+++ b/src/synthesizer/particle/stars.py
@@ -1973,16 +1973,30 @@ class Stars(Particles, StarsComponent):
             label=label,
         )
 
-    def calculate_integrated_spectra(self):
+    def calculate_integrated_spectra(self, update=True):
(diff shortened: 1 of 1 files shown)


WillJRoper:
This already exists: https://github.com/flaresimulations/synthesizer/blob/main/src/synthesizer/particle/galaxy.py#L423C1-L449C17

Please check.

WillJRoper:
If you wanted to include this on a component you could take the loop from that method and move it to the `Particles` base class and then wrap that in the method above.

stephenmwilkins:
Ah... so it does, I was only looking on the component. I think it's useful on component but maybe not essential.

WillJRoper:
Superseded by #635

PR #619, opened 2024-05-14
Raise warning when particle arrays empty and setting stars or gas to none

christopherlovell:
When loading stars or gas particles into a stars or gas component, previously if one of these arrays was not present (or spelled incorrectly) it would silently just return `None`. This adds an explicit warning, that can be disabled by setting verbosity to False.

## Issue Type
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/particle/galaxy.py
+++ b/src/synthesizer/particle/galaxy.py
@@ -191,6 +191,8 @@ class Galaxy(BaseGalaxy):
                 Star particle metallicity (total metal fraction)
             stars (stars particle object)
                 A pre-existing stars particle object to use. Defaults to None.
+            verbose (bool)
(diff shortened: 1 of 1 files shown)


src/synthesizer/particle/galaxy.py
```diff
@@ -206,6 +207,12 @@ def load_stars(
                 | (ages is None)
                 | (metallicities is None)
             ):
+                if verbose:
+                    print(
+                        ("In `load_stars`: one of either `initial_masses`"
```
WillJRoper, line 212: For the record these inner wrapping brackets are unnecessary because the print already wraps the strings.

WillJRoper (approved):

PR #635, opened 2024-05-28
Moved `intergrate_particle_spectra` to component level

WillJRoper:
Supercedes #618.

This PR moves the method for integrating all particle spectra from solely on a galaxy to the components. This allows a user only working with components to quickly integrate all spectra. 

`Galaxy.integrate_particle_spectra` is now just a wrapper around the component versions.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/particle/particles.py
+++ b/src/synthesizer/particle/particles.py
@@ -227,7 +227,7 @@ class Particles:
 
         return self.particle_photo_fluxes
 
-    def integate_particle_spectra(self):
+    def intergate_particle_spectra(self):
         """
         Integrate any particle spectra to get integrated spectra.
 
--- a/src/synthesizer/particle/particles.py
+++ b/src/synthesizer/particle/particles.py
@@ -227,7 +227,7 @@ class Particles:
 
         return self.particle_photo_fluxes
 
-    def intergate_particle_spectra(self):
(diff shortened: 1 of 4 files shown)


src/synthesizer/particle/particles.py
```diff
@@ -227,6 +227,18 @@ def get_particle_photo_fluxes(self, filters, verbose=True):
...
+        """
+        Integrate any particle spectra to get integrated spectra.
+
+        This will take all spectra in self.particle_spectra and call the sum
+        method on them, populating self.spectra with the results.
+        """
+        # Loop over the particle spectra
+        for key, sed in self.particle_spectra.items():
+            # Sum the spectra
+            self.spectra[key] = sed.sum()
```
christopherlovell, line 240: I assume this sum is overloaded to ensure it's acting over the correct dimensions?

WillJRoper: Indeed, see https://github.com/flaresimulations/synthesizer/blob/main/src/synthesizer/sed.py#L145 

However, now you mention it this will only ever sum over the first dimension. We never do currently but if we had a ndim > 2 `sed` this wouldn't behave as expected. We want it to sum over all but the last axis.

christopherlovell (approved):

PR #636, opened 2024-05-28
Converted prints to warnings

WillJRoper:
We've had a lot of rogue prints for a while now and they have led to large logs when running pipelines. Some of these were wrapped by a verbose flag but that was a half way house solution.

This PR introduces a wrapper around `warnings.warn` which now replaces all these rogue prints (only the prints which correspond to warning the user of something). 

The reason I have used a wrapper rather than just calling `warnings.warn` is so we can overload the default category and stack level instead of stating these in each warning. 

In the future we could expand on the default `RuntimeWarning` with specific Synthesizer warnings.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/dust/emission.py
+++ b/src/synthesizer/dust/emission.py
@@ -26,6 +26,7 @@ from unyt.dimensions import temperature as temperature_dim
 from synthesizer import exceptions
 from synthesizer.sed import Sed
 from synthesizer.utils import planck
+from synthesizer.warnings import warn
 
 
 class EmissionBase:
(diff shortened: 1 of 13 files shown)


src/synthesizer/dust/emission.py
```diff
@@ -450,7 +447,8 @@ def get_spectra(self, _lam, dust_components=False):
         """
 
         if self.template == "DL07":
-            print("Using the Draine & Li 2007 dust models")
+            if verbose:
```
christopherlovell, line 450: should this be a warn?

src/synthesizer/sed.py
```diff
@@ -965,11 +966,9 @@ def get_photo_luminosities(self, filters, verbose=True):
             # and the spectral grid are the same array
             if not np.array_equal(f.lam, self.lam):
                 if verbose:
```
christopherlovell, line 968: should this verbose flag still be here?

src/synthesizer/dust/emission.py
WillJRoper: I spent a while thinking about that but it's more of a "statement of fact" than a warning hence why I placed it in a verbose check.

src/synthesizer/sed.py
WillJRoper: Absolutely not! It's been obliterated with maximum prejudice.

christopherlovell (approved):

PR #637, opened 2024-05-28
Implementing errors and malloc checks in C extensions, improving setup.py

WillJRoper:
The C extensions have always errored extremely opaquely. This is mainly because I didn't know how to error at the interface between C and Python when I wrote them. That has thankfully changed!

I've now implemented proper errors everywhere the extensions would return `NULL` but in addition I have also put in checks for every extraction from a Python object to a C array, and importantly every malloc is now followed by a check that the malloc was successful. This means the memory errors prevalent when running high resolution particle spectra will error telling the user exactly which malloc caused the error (and also signalling clearly that it is a memory related error not a SEGFAULT).

In addition to the above I have also rewritten `setup.py` to be more "modern" and allow optional setting of `CFLAGS` and `LDFLAGS` for testing. This also lays the groundwork for more complex builds if we want to use OpenMP etc. without putting the complications onto the inexperienced user. 

These changes MAY also make Synthesizer work with Windows.

I have added some details about the new installation options to the docs but flagged that it's unimportant for most.

## Issue Type

- Enhancement

## Checklist


WillJRoper:
Currently, this is broken due to changes I was making to handle compilation errors observed on Windows. Switching it to a draft until they're fixed.

WillJRoper:
MVSC is still proving stubborn. I can't fathom the errors it's triggering on but I don't have easy access to a windows computer to test on 😅

WillJRoper:
I think I've solved the issues MVSC complains about. If new complaints come up we can look at that on a future PR. 

This PR runs on all OSs Synthesizer previously ran on and adds some really valuable safety aspects which would be useful for some users currently debugging issues. Therefore I think it's worth getting this one in.

aswinpvijayan (approved):

PR #644, opened 2024-06-05
Los orientation

aswinpvijayan:
The current implementation of the los calculation is in the -z direction as opposed to +z, which was giving a different result from previous FLARES calculation. This is not necessarily an issue, but for consistency sake I have switched to the +z direction. 

Also increased the resolution of the default kernel from 1000 -> 10000 bins as we had in FLARES.

## Checklist


WillJRoper:
Doesn't -z make more intuitive sense? Placing the "observer" at 0?

WillJRoper (approved):

PR #646, opened 2024-06-10
Adds a plotting script for interfacing with valgrind and profiling scripts

WillJRoper:
DWISOTT. 

## Issue Type

- Enhancement

## Checklist
- [x] I have read the [CONTRIBUTING.md]() -->
- [x] I have added docstrings to all methods
- [x] I have added sufficient comments to all lines
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no pep8 errors
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #647, opened 2024-06-10
Some much needed clean up and abstraction in the extensions

WillJRoper:
This PR cleans up the spectra extraction C extensions. In doing so it introduces a lot of helper functions and structs to reduce the boilerplate and number of arguments handed around respectively.

Some of the refactoring enables the workflow necessary for openMP threading.

This branch contains the scripts from #646, so that should be merged first.

## Issue Type

- Enhancement

## Checklist


WillJRoper:
This did start life as an attempt to use a hash map for the weights (and thus spectra) to reduce the memory footprint. In practice, the hash map added a lot of overhead slowing down the computation and the reduction in memory paled in comparison to other objects (e.g. particles, the spectra grid etc.) meaning it was an overall negative.

WillJRoper:
Note that the results of these changes have been checked against the existing docs and give the same answers.

christopherlovell (approved):

PR #648, opened 2024-06-11
Introduced internal shared memory parallelism in spectra extensions with OpenMP

WillJRoper:
This PR introduces OpenMP threading in the spectra generation extensions.

- OpenMP can now be optionally installed. 
- An extension to test if OpenMP is installed has been introduced (using a preprocessor directive).
- A test has been introduced to check the OpenMP installation worked in the workflow.
- Integrated lines and spectra can now use threading to calculate the weights and then perform the summing of the weights convolved with the spectra grid.
- Particle lines and spectra can now use threading to calculate the spectra. (This is embarrassingly parallel but in reality there's always a price to pay when threads queue up to read the same data).
- A timing extension can now be turned on using the `ATOMIC_TIMING` configuration flag which will print thread-safe timing of different operations. This has been implemented in a few places both in the C and Python for profiling purposes.

## Issue Type

- Enhancement

## Checklist


WillJRoper:
hierarchical_integrated_strong_scaling_cic_NStars1000000_TotThreahs8
This shows the scaling of integrated spectra. It scales relatively well up until the limit where the python overhead catches up. Mainly the bolometric luminosity calculation.

WillJRoper:
no_allocate_particle_strong_scaling_cic_NStars50000_TotThreahs8
Particle spectra are a similar (but worse) story. The red line here is the parallel region of the code. It doesn't scale incredibly well but "well enough" I suspect there's some cache locality overhead which will be difficult to get around. However, that at least scales! The big issues here is the bolometric luminosity calculation. 

This calculation always uses quad for particle spectra which is overkill and inefficient. This should be solved in #544 so maybe it's worth getting that done and dusted before rerunning this.

WillJRoper:
Note that SFZH, and both lines behave the same as the spectra but are significantly cheaper.

aswinpvijayan (approved):

PR #650, opened 2024-06-17
Added Swift-EAGLE loader

christopherlovell:
Added loader for Swift-EAGLE simulations in the CAMELS suite.

**Important**: introduces a *breaking change* for CAMELS data loaders. Standardises the naming of snapshot and group files across the different simulation data loaders, so rather than `fof_name` and `fof_dir` these are now `group_name` and `group_dir` across all methods.     

*Note*:
Unfortunately we don't have particle data files, so the particle load from the raw snapshots is a big bottleneck. Some embarrassingly parallel multithreading (through `schwimmbad`, for example) would speed things up a lot. 

## Issue Type
- Enhancement

## Checklist


christopherlovell:
Changes moved to new PR post-emission models merge

PR #661, opened 2024-07-12
Enabling the discarding of spectra for memory savings

### LINKED ISSUE #659: `EmissionModels` should have a `save` attribute
Not all emisisons produced by a model need to be saved (i.e. attached to an emitter and held on to). To enable the "forgetting" of spectra (and thus not wasting memory) we should allow an `EmissionModel` to define whether its spectra should be saved by introducing a `save` attribute which can be `True` by default.

WillJRoper:
This PR closes #659 by introducing a `save` flag on an `EmissionModel` to enable the discarding of spectra rather than keeping them attached to an emitter.

## Issue Type

- Enhancement

## Checklist


aswinpvijayan:
Why is there a need to delete lines which do not have a saev flag? Seems a bit overkill since the user usually requests the lines?

WillJRoper:
> Why is there a need to delete lines which do not have a saev flag? Seems a bit overkill since the user usually requests the lines?

It's not the lines that are being discarded. It's the "emission key", i.e. all lines will always appear for each emission key that is saved. Unlike spectra, lines are selected at the call to `get_lines`.

aswinpvijayan (approved):

PR #678, opened 2024-07-23
Fix doc build example failures

christopherlovell:
Fixes examples so that (a) during the docs build, if they fail, the build fails, and (b) that they are updated to emission models, as these were missed in the original switch over because of (a)

## Issue Type
- Bug

## Checklist


docs/source/conf.py
```diff
@@ -50,13 +50,11 @@
...
-    # Modules for which function/class level galleries are created. In
-    # this case sphinx_gallery and numpy in a tuple of strings.
+    # Modules for which function/class level galleries are created
     "doc_module": ("synthesizer"),
-    #     # Regexes to match objects to exclude from implicit backreferences.
-    #     # The default option is an empty set, i.e. exclude nothing.
-    #     # To exclude everything, use: '.*'
-    #     # 'exclude_implicit_doc': {r'pyplot\.show'},
+    "abort_on_example_error": True,
+    # "show_memory": True,
```
WillJRoper, line 56: Any reason to leave these here commented?

WillJRoper: If show memory shows the memory usage in the logs that sounds useful!

WillJRoper:
Looks good to me, what's left to do on this?

christopherlovell:
Still a lot, none of the cosmo examples work presently. Please hold :)

docs/source/conf.py
christopherlovell: I thought about it but it's expensive. I think okay to leave here for users to comment out when testing locally?

WillJRoper (approved):

Issue #608, opened 2024-05-13
Replace calculate_los_tau_v with generic column density method


stephenmwilkins:
At the moment we have the method `calculate_los_tau_v` on galaxy; this calculates the column density of metals, converts to a dust mass, and then calculates an optical depth (`tau_v`). 

In addition to being fairly opaque, I think we would be better served by having a more generic `calculate_column_density` method. This should be flexible enough to calculate both the column density of gas but also the column density of any property of the gas (e.g. metals, a specific element, or dust if it exists).

We could then have a method which wraps this to calculate the dust column depth and then a method that wraps that to calculate the optical depth. However, these could also be just part of a pipeline?

WillJRoper:
It would be nice to have a generic column density method but surely that would would better be done via the map functionality already implemented?

`calculate_los_tau_v` is specifically relative to each star (to get `tau_v` for each star), I'm not sure it makes sense to calculate any other column density for each star. A lesser point, but if I remember correctly there were specifics that had to be done in the C making it a bit more difficult to generalise.

What is the use case for any other column density specifically relative to the position of each star?

WillJRoper:
In fact, I think replacing this with a generic `calculate_column_density` method is confusing. I would expect a generic column density not to be relative to a star and instead be through the whole galaxy producing a map. Lets not change `calculate_los_tau_v`.

stephenmwilkins:
Ah... so there is distinction, the entire column density of the galaxy, which I agree can be done on the map, or the column density to each star particle, which is this calculation. I think they just need a different name but we definitely want a generic column density - to each star - method. It would be useful to know how much gas, metals, and dust (when dust is a property of the gas, as in Simba and COLIBRE) each star particle is seen through.

`calculate_stellar_column_density` implies the column density of stars so we can't use that. How about `calculate_column_density_to_stars`? or `calculate_column_density_to_star_particles`.

WillJRoper:
Or we could attach it to a stellar component instead and have it take a gas object. That makes sense I think.

stephenmwilkins:
As for use case:

- using `metallicity` is important for the current LOS dust calculation
- using a specific element (e.g. carbon or silicon) may be useful for a more sophisticated LOS dust calculation.
- for models with dust as a property of the gas particle obviously we should use that instead for the LOS dust calculation.
- using just the mass of hydrogen (functionally ~75% of the gas mass) could be used to calculate an upper estimate of the ionising photon escape fraction.

[...]

PR #684, opened 2024-07-24
Optimisation and refactor of LOS calculations

WillJRoper:
This PR implements a rewrite of the LOS calculation. Specifically:

- Previously only los optical depths were implemented, now arbitrary column densities can be calculated too (closing #608).
- `calculate_los_tau_v` is now deprecated in favour of `get_<component>_los_tau_v` methods to that BH tau_vs can be calculated with a nice interface. The original method has been kept for now for backwards compatibility.
- Line of sight calculations are now documented.
- The extensions for the LOS calculation have been optimised now for vastly improved tree performance and better decision for when a simple loop is possible. In addition, some arguments have now been moved to the frontend to enable tuning of the tree in the rare instance tuning is necessary.
- LOS calculation can now use threading.

The improved tree performance yields a much bigger speed up relative to a simple loop:

the_end_of_suffering

While the code scales extremely well:

los_strong_scaling

## Issue Type

- Document
- Enhancement

## Checklist


christopherlovell (changes requested):
As discussed on slack, just need to simplify example and move some content to a dedicated profiling script. Other than that this is ready

PR #692, opened 2024-07-29
Small fixes

christopherlovell:
- Small fixes to Swift-EAGLE CAMELS load method.
- Creates a `log10metallicities` property on gas components.
- Fixes masking errors when using apertures, parametric_young_stars and other masks simultaneously 

## Issue Type
- Bug

## Checklist


WillJRoper:
Should we just move the log10 properties to the `Particles` base class?

src/synthesizer/particle/gas.py
```diff
@@ -140,7 +140,7 @@ def __init__(
 
         # Set the metallicites and log10 equivalent
         self.metallicities = metallicities
-        self.log10metallicities = np.log10(self.metallicities)
+        # self.log10metallicities = np.log10(self.metallicities)
```
WillJRoper, line 143: Is there a reason this has been left?

src/synthesizer/load_data/load_camels.py
```diff
@@ -676,6 +678,8 @@ def swifteagle_particle_assignment(
     ):
         gal = Galaxy()
         gal.redshift = redshift
+        gal.centre = pos[idx] * kpc
```
WillJRoper, line 681: Just checking but definitely kpc here? The rest appears to be in Mpc

src/synthesizer/particle/stars.py
```diff
@@ -612,6 +594,11 @@ def _aperture_mask(self, aperture_radius):
                 Radius of spherical aperture in kpc
         """
 
+        if self.centre is None:
+            raise ValueError(
+                "Centre of star particles must be set to use aperture mask."
+            )
+
         distance = np.sqrt(
```
WillJRoper, line 602: There's a property attached to all particles called something like `centered_coordinates` which automatically hands the centered coordinates and also ensures centre has been set at the same time. 

You could remove the check by using that property here.

WillJRoper (changes requested):
A couple checks and one minor simplification.

src/synthesizer/particle/gas.py
christopherlovell: No, removed

src/synthesizer/load_data/load_camels.py
WillJRoper: That's tragic 😂

src/synthesizer/particle/stars.py
christopherlovell: moved to particle parent class, and simplified to use existing methods

WillJRoper (approved):

PR #696, opened 2024-08-01
Replace equivalent_width attribute with method

### LINKED ISSUE #590: Add ability to recalculate equivalent widths
At the moment the equivalent width (EW) of a line is set when the `Line` object is initialised.  The methods that create `Line` objects for stars do this automatically by combining the various continuum sources. However, it might be useful to have a method on `Line` (and wrapper on `LineCollection`) that allows the continuum, and thus equivalent width, to be updated. 

This might also be useful for more complicated arrangements like we'll have for AGN.

stephenmwilkins:
At present the `equivalent_width` is calculated on initialisation of a `Line` object. However, if the line or continuum luminosity is updated it will then be wrong. 

Instead I've replaced the attribute with a method and added the `@property` decorator. 

Closes #590.  

## Checklist

### DIFF
--- a/src/synthesizer/line.py
+++ b/src/synthesizer/line.py
@@ -651,14 +651,14 @@ class Line:
         self.element = [li.strip().split(" ")[0] for li in self.id.split(",")]
 
     @property
-    def continuum_lam(self):
+    def continuum_llam(self):
         """Return the continuum in units of Llam (erg/s/AA)."""
         return lnu_to_llam(self.wavelength, self.continuum)
 
     @property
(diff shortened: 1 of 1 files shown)


WillJRoper (approved):
Looks good and definitely a better implementation since the EW is so cheap to calculate. To save you some time I fixed the only comments I had:
- Single sentence doc strings should be on 1 line (my editor will complain at me if they are not and it is PEP-8).
- `continuum_lam` -> `continuum_llam` because it's clearer.

PR #698, opened 2024-08-06
Age grid updates and check added

christopherlovell:
Fixes a small interpolation issue by locking the base of the grid on the scale factor of the snapshot being loaded.

Adds a check and error when loading ages into a particle stars object if those ages are negative.

## Issue Type
- Bug
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/load_data/utils.py
+++ b/src/synthesizer/load_data/utils.py
@@ -28,13 +28,19 @@ def get_len(Length):
     return begin, end
 
 
-def age_lookup_table(cosmo, delta_a=1e-3, low_lim=1e-4):
+def age_lookup_table(cosmo, redshift=0., delta_a=1e-3, low_lim=1e-4):
     """
-    Create a look-up table for age as a function of scale factor
+    Create a look-up table for age as a function of scale factor.
+
+    Defaults to start at the lower resolution limit (`delta_a`),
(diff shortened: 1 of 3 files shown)


src/synthesizer/load_data/utils.py
```diff
@@ -28,13 +28,19 @@ def get_len(Length):
     return begin, end
 
 
-def age_lookup_table(cosmo, delta_a=1e-3, low_lim=1e-4):
+def age_lookup_table(cosmo, redshift=0., delta_a=1e-3, low_lim=1e-4):
```
WillJRoper, line 31: Should redshift not be a positional argument? There will always be an associated snapshot with a redshift.

WillJRoper (approved):
All good but just one question about the redshift argument. Happy to merge as is though if you don't think it should be a positional argument.

src/synthesizer/load_data/utils.py
christopherlovell: I agonised over this a bit. Settled on this as I wanted to make it as general as possible. The sim may be a lightcone, so you may just want to produce a lookup grid at all times. You may also wish to produce a single lookup grid for a sim for all snapshots, to further reduce compute. With this you can do this, without having to explain a default

WillJRoper: All good points! Fair play.

PR #700, opened 2024-08-09
Update parametric SFH models

stephenmwilkins:
Updates the parametric SFH models, removing redundancy and generalising the models by allowing all models to take a `min_age` and `max_age` parameter. Also changed the definition of `tau` (by swapping the sign) to match the standard in the literature. 

## Issue Type

- [x] Enhancement

## Checklist


src/synthesizer/parametric/sf_hist.py
```diff
@@ -232,81 +240,50 @@ def _sfr(self, age):
...
         max_age (unyt_quantity)
-            The age at which the exponential is truncated. Above this age
-            the SFR is 0.
+            The age above which the star formation history is truncated.
+        min_age (unyt_quantity)
+            The age below which the star formation history is truncated.
     """
 
-    def __init__(self, tau, max_age):
+    def __init__(self, tau, max_age=1e11 * yr, min_age=0.0 * yr):
```
aswinpvijayan, line 261: Maybe the maximum age could be set to something reasonable as default

src/synthesizer/parametric/sf_hist.py
```diff
@@ -316,11 +293,16 @@ def _sfr(self, age):
...
         """
-        if age < self.max_age:
+
+        if (age < self.max_age) and (age > self.min_age):
             return np.exp(-age / self.tau)
         return 0.0
 
 
+# included for backwards compatability
+TruncatedExponential = Exponential
```
aswinpvijayan, line 303: Let's just remove this truncated completely

aswinpvijayan:
Maybe 100 Gyr is a bit too much

src/synthesizer/parametric/sf_hist.py
```diff
@@ -383,147 +375,59 @@ def _sfr(self, age):
         return 0.0
 
 
-class ExponentiallyDeclining(Common):
```
aswinpvijayan, line 386: We still do need these

aswinpvijayan (changes requested):

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: Possibly, though in most cases it won't make a difference at all.

stephenmwilkins:
Incidentally, I don't think the previous implementation of DelayedExponential really was delayed, but truncated?

src/synthesizer/parametric/sf_hist.py
aswinpvijayan: Isnt this the exponential form: np.exp(-(self.max_age-age) / self.tau)

stephenmwilkins: I checked, this is equivalent to what I've done but swapping the sign of `\tau`. We could implement it that way or simply swap the sign of `\tau` if that allows us to align with other conventions?

src/synthesizer/parametric/sf_hist.py
WillJRoper: Why are these being given defaults at all? At least for `max_age`. This is a important parameter of the model. It's totally reasonable you'd expect the user to always give it.

src/synthesizer/parametric/sf_hist.py
WillJRoper: Don't remove it just yet but don't do this either. 

You should make a class that just calls `Exponential.__init__` in it's own `init`. You can then add the deprecated decorator to `TruncatedExponential.__init__` which will warn that it will be removed. 

e.g.
```
class TruncatedExponential(Exponential):
    @deprecated
    def __init__(self, *args, **kwargs):
        Exponential.__init__(self, *args, **kwargs)
```

src/synthesizer/parametric/sf_hist.py
WillJRoper: I'd keep both. It's nice sign posting. Makes it very clear to the user what to use if they want a certain thing. Multi-use classes/functions just add to confusion. 

By all means have one inherit from the other and swap the sign on `tau` to reduce boilerplate but we 100% should leave these classes available.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: I somewhat agree with `max_age`, and forcing its use makes people consider the age of the Universe in their particular scenario. changing `min_age` from 0 fundamentally changes the model though.

WillJRoper: Sure, `min_age` I get. I mainly mean `max_age`. It works better as a positional argument and means you don't need to muck about with defaults and/or `Nones`

src/synthesizer/parametric/sf_hist.py
```diff
@@ -118,9 +120,29 @@ def calculate_sfh(self, t_range=(0, 10**10), dt=10**6):
...
+
+        Args:
+            t_range (tuple, float)
+                The age limits over which to calculate the SFH.
+            dt (float)
+                The interval between age bins.
+        """
+
+        t, sfr = self.calculate_sfh(t_range=t_range, dt=dt)
+        plt.plot(t, sfr)
```
WillJRoper, line 136: Would this not be better done as a step plot given the gridded nature of the SFH?

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: I disagree, I think a single model makes more sense. We just need to decide on our `\tau` convention and be clear. Why have two models when one will suffice? Especially since you might like to explore a range of \tau encompassing positive and negative values.

Another part of the issue is that "Declining" and "Increasing" are a little ambiguous. In common usage a "DecliningExponential" means a galaxy whose recent star formation is less than it previously was. However, since we use age instead of time as our parameter a "DecliningExponential" is actually increasing SFR with age.

src/synthesizer/parametric/sf_hist.py
```diff
@@ -17,6 +17,9 @@
 
 """
 
+from warnings import warn
```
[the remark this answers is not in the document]
WillJRoper: i.e. there's a warnings module which includes deprecation warnings explictly, use it.

src/synthesizer/parametric/sf_hist.py
```diff
@@ -199,28 +221,46 @@ class Constant(Common):
...
+                If min_age = 0 then this is the duration of star formation.
+            min_age (unyt_quantity)
+                The age below which the star formation history is truncated.
             duration (unyt_quantity)
-                The duration of the period of constant star formation.
+
         """
 
+        if duration is not None:
+            warn(
```
WillJRoper, line 250: As above, there is a deprecation function built into Sythesizer.

src/synthesizer/parametric/sf_hist.py
```diff
@@ -199,28 +221,46 @@ class Constant(Common):
...
+                "The use of duration is deprecated in favour of max_age",
+                DeprecationWarning,
+                stacklevel=2,
```
WillJRoper, line 253: Stack level 2 here wouldn't report the right level anyway.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: Yeah, done.

src/synthesizer/parametric/sf_hist.py
```diff
@@ -383,147 +466,59 @@ def _sfr(self, age):
...
-        Args:
-            age (float)
-                The age (in years) at which to evaluate the SFR.
-        """
-        if age < self.initial_age:
-            return np.exp(-1 * (self.initial_age - age) / self.tau)
-        return 0.0
-
-
-class DelayedExponentiallyDeclining(Common):
```
WillJRoper, line 435: Same as in the comment above, we should keep these. Much simpler for the user to have obvious naming signposting them. If they are just wrappers then that is fine.

WillJRoper (changes requested):

src/synthesizer/parametric/sf_hist.py
WillJRoper: I'm not going to capitulate on this one specifically. Keep both.

> Why have two models when one will suffice? 
Because it is immediately obvious to the user without them needing to think about what conventions we chose. They just pass a number. 

> Especially since you might like to explore a range of \tau encompassing positive and negative values.
That can be done with the base model regardless.

I agree it's pointless having all the redundant definitions if they can all just stem off a base class but the labelling alone makes it worthwhile.

> Another part of the issue is that "Declining" and "Increasing" are a little ambiguous. In common usage a "DecliningExponential" means a galaxy whose recent star formation is less than it previously was. However, since we use age instead of time as our parameter a "DecliningExponential" is actually increasing SFR with age.
That strikes me as a big problem in itself. We should follow what you might expect (as you describe). With named classes that just take a `tau` that's easily done with the specific naming because the user just passes a `tau` and should get a declining SFH (declining from a peak in the past). 
If you instead use a one size fits all method they need to know our convention, know it is backwards and then pass the correct `tau` following our choices.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: Not here right, `calculate_sfh` doesn't integrate across a bin but gets the SFR at the times given.

docs/source/components/stars.ipynb
```diff
@@ -366,7 +366,7 @@
     "print(metal_dist)\n",
     "\n",
     "# Define a constant SFH\n",
-    "sfh = SFH.Constant(duration=100 * Myr)\n",
```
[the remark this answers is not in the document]
WillJRoper: I see you're just defining the start and end points now. That makes sense and is more intuitive. You could add the option for any combination of max, min and duration but probably just complicates things for the sake of it.

src/synthesizer/parametric/sf_hist.py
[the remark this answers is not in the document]
stephenmwilkins: Done

src/synthesizer/parametric/sf_hist.py
WillJRoper: Although... the same argument stands here. We could keep `TruncatedExponential` as a nice wrapper and have the max and min arguments as positional (i.e. they are needed for a truncated exponential SFH).

docs/source/components/stars.ipynb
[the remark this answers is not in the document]
stephenmwilkins: If we only had `Constant` I'd use `duration` and `min_age` (i.e. `max_age = min_age + duration`), with `min_age` by default being 0. However, I think it's better to harmonise with the other models and use `min_age` and `max_age`. `max_age` is more useful in those models because you'd probably naturally set it to the age of the Universe, or some large fraction of it.

I did ponder allowing all three but then you'd have no positional arguments and you'd have to add various logic. 

As annoying as this change is, I think it is better.

src/synthesizer/parametric/sf_hist.py
WillJRoper: Ah yea, forgot that.

docs/source/components/stars.ipynb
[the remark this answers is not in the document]
WillJRoper: Yea, on balance I think I agree.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: I have harmonised tau with the standard definition where tau is positive for a declining exponential. I've also added a DecliningExponential where tau is forced to be positive.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: The delayed model was broken, it's now implemented properly as `DelayedExponential`.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: Done.

src/synthesizer/parametric/sf_hist.py
stephenmwilkins: Done.

examples/parametric/compare_sfhs.py
```diff
@@ -0,0 +1,55 @@
+import matplotlib.pyplot as plt
```
christopherlovell, line 1: In order for an example to show a plot in the examples gallery it needs to follow a specific naming and formatting convention. See the [contributing guidelines](https://github.com/flaresimulations/synthesizer/blob/main/docs/CONTRIBUTING.md#adding-example-scripts) for details.

christopherlovell (changes requested):

src/synthesizer/parametric/sf_hist.py
```diff
@@ -17,19 +17,20 @@
...
 # Define a list of the available parametrisations
 parametrisations = (
     "Constant",
-    "Exponential" "TruncatedExponential",
+    "Gaussian",
+    "Exponential",
     "LogNormal",
-    "ExponentiallyyDeclining",
-    "DelayedExponentiallyDeclining",
     "DoublePowerLaw",
```
aswinpvijayan, line 34: Add the DelayedExponential here?

examples/parametric/plot_compare_sfhs.py
```diff
@@ -0,0 +1,62 @@
+"""
+Compare different parametric star formation history models
+===========================
```
WillJRoper, line 3: Equals need to be the same length as the title. It will fail typesetting otherwise.

src/synthesizer/parametric/sf_hist.py
WillJRoper: There's a few missing actually

WillJRoper:
Only a few more things to clean up now and then I think this is good to go. Can you update the checkboxes and the TODO list in the description @stephenmwilkins. I assume the TODOs are no longer applicable since you opened the PR?

stephenmwilkins:
All done I think

WillJRoper (approved):
All looks good to me now.

aswinpvijayan (approved):

PR #704, opened 2024-08-13
Clearer photometry naming convention

WillJRoper:
DWISOTT

## Issue Type

- Enhancement

## Checklist


WillJRoper:
As mentioned on slack, theres an issue here.

I can change the `PhotometryCollection` naming convention without issue because it's backend but `Galaxy` defines front facing attributes for `photo_fluxes` and the methods also use this naming convention (same is true for the luminosity equivalents). Changing these would be a front-end change.

WillJRoper:
I fixed the above issue with properties and wrappers to maintain old behaviour (with deprecation warnings). Everything uses the new naming convention but the old one will work.

aswinpvijayan (approved):

PR #705, opened 2024-08-14
Introducing an aperture method for an `Image`

WillJRoper:
This PR introduces a method for applying circular apertures to an `Image` to get the sum of the "signal" in the aperture. This adapts the method used in photutils but introduces threading (because I can) and is between 5 and 10 times faster (in serial) than the photutils version for getting the "aperture photometry" (as they call it). 

An example has been introduced that shows the calculation reproduces the theoretical result exactly. I have not included it in the docs because this is a sufficiently niche use case. If we add any more image analysis then maybe there's an argument for a docs section.

## Issue Type

- Enhancement

## Checklist


src/synthesizer/imaging/extensions/circular_aperture.c
```diff
@@ -0,0 +1,422 @@
...
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+#include <numpy/arrayobject.h>
+
+/* Local includes */
+#include "../../extensions/property_funcs.h"
+#include "../../extensions/timers.h"
+
+/* It's possible we don't have PI... if so define it. */
+#ifndef M_PI
+#define M_PI 3.141592653589793238462643383279502884
```
aswinpvijayan, line 30: Do we need that much precision?

aswinpvijayan (approved):

src/synthesizer/imaging/extensions/circular_aperture.c
WillJRoper: Why not ;). To be fair, this will rarely be used since it'll be defined most of the time.

PR #708, opened 2024-08-16
Introduced rotation methods for particles and galaxies

### LINKED ISSUE #12: Need to include geometry change helper functions
Where should we put these? Things like rotation matrices, face-on, edge-on. Possibly in the galaxy class?

WillJRoper:
This PR introduces the ability to rotate a galaxy/component either in place or returning a new instance. An example demonstrating it has been introduced. 

At the galaxy level the user can select which component's angular momentum should be used for the vector to rotate towards. 

To make this work an angular momentum property has been introduced on a `Particles` object.

Close #12 (finally).

## Issue Type

- Enhancement

## Checklist


WillJRoper:
rot_example

This is the image the example produces demonstrating rotating to face on and edge on based on stellar and gas angular momentum

src/synthesizer/utils/geometry.py
```diff
@@ -0,0 +1,69 @@
...
+    unit = np.array([0.0, 1.0, 0.0])
+
+    # Get the rotation matrix
+    matrix = get_rotation_matrix(vector, unit)
+"""
+
+import numpy as np
+
+
+def get_rotation_matrix(vector, unit=None):
```
aswinpvijayan, line 21: The method is taken directly from stackoverflow, would be nice to quote that: https://stackoverflow.com/questions/43507491/imprecision-with-rotation-matrix-to-align-a-vector-to-an-axis

WillJRoper: Included

WillJRoper:
There has been a significant ruff update that meant everything was breaking (they have improved the import sorting from what I can tell, it now appears to automatically separate internal and external imports).

I've updated the workflows and pre-commit hook to account for this but it means there's a ton of formatting.

aswinpvijayan (approved):

PR #717, opened 2024-08-27
Handle sed objects with ndim > 2

christopherlovell:
Grids have ndim > 2 (e.g. age, metallicity, wavelength). Some of the `sed` class methods strictly test for 2 dimensions, which leads them to break. This updates those checks, to handle sed objects with higher numbers of dimensions.

## Issue Type
- Bug

## Checklist

### DIFF
--- a/src/synthesizer/sed.py
+++ b/src/synthesizer/sed.py
@@ -610,7 +610,7 @@ class Sed:
         # Apply the correct method
         if integration_method == "average":
             # Apply to the correct axis of the spectra
-            if self.ndim == 2:
+            if self.ndim >= 2:
                 lnu = (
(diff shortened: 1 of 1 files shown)


WillJRoper (commented):
I've just set the tests running after the download fail. Happy to approve once they pass but I've mulled these over before... surely there's a way of doing this while avoiding the dimension logic entirely 🤔

christopherlovell:
Yeah I did wonder that, but unfortunately a lot of these methods rely on integration routines from scipy that are necessarily 1D.

This fixes the immediate issue for now at least.

WillJRoper (approved):

PR #718, opened 2024-08-28
Fix parametric young stars

christopherlovell:
Updates the parametric young stars implementation to create individual star particles on the grid. This is opposed to the previous implementation, which was implemented on `generate_lnu` and directly produced the spectra from a parametric stars object; this led to issue with emission models with age dependencies, e.g., the CF00 dust model.

In this implementation new stars are created on the grid. These can then be self consistently treated as normal star particles for all downstream tasks. The old stars are saved on the object, as well as details of the parametrisation. This means that the procedure can be reversed, or new parametrisations applied, relatively easily.

This relies on the `add_parts` branch, and makes some additional modifications to this to generalise addition of particle stars object array attributes. 

Additional changes:
- renames `particle.stars.sample_sfhz` to `sample_sfzh`
- fixes a bug in generate_lnu where `fesc` could be equal to None
- adds checks for nan or inf on particle stars input (ages, metallicities, initial_masses)
- adds properties on particle stars object for integrated total mass and SFR (optional timescale)
- fixes shape issues in some `measure_` methods on `sed`

## Issue Type
- Enhancement
- Bug

## Checklist


WillJRoper (approved):

examples/cosmo/plot_parametric_young_stars.py
WillJRoper, line 1: I've edited this to make it run but I'm not sure it's actually giving the right answer. Could you take a look and make sure you're happy with the output? Currently it outputs this plot:
resample_example
Which doesn't look at all right.

WillJRoper (changes requested):
This looks good overall but possibly some sticky corners still.

Currently we're missing a method to "unresample", i.e. it would be nice to have a front facing function that will return a `Stars` object to its pre-resampling state using `_remove_stars`.

src/synthesizer/particle/stars.py
```diff
@@ -590,29 +839,66 @@ def generate_lnu(
...
 
-    def _parametric_young_stars(
+        # Check the arguments we've been given
+        self._check_star_args()
+
+    def parametric_young_stars(
         self,
-        pmask,
         age,
         parametric_sfh,
```
WillJRoper, line 890: Can we not request a class (uninstatiated) here in stead of a string? Feels a bit more intuitive and future proof.

src/synthesizer/particle/stars.py
```diff
@@ -641,8 +984,57 @@ def _parametric_young_stars(
...
+            ]
+        )
+
+        # Find the grid indexes on the parametric grid
+        grid_indexes = index_pairs[stars.sfzh > 0]
+
+        # Create new particle stars object from non-empty SFZH entries
+        new_stars = self.__class__(
+            stars.sfzh[stars.sfzh > 0] * Msun,
+            10 ** grid.log10ages[grid_indexes[:, 0]] * yr,
```
WillJRoper, line 1003: FYI @christopherlovell, found a bug here. This was previously `grid.log10ages[grid_indexes[:, 0]]` but particle stars take pure ages.

WillJRoper (approved):

PR #725, opened 2024-09-06
Ensuring unit consistency in Emission generators

WillJRoper:
While ensuring the dust emission being generated was actually right I stumbled across some inconsistencies. To avoid these inconsistencies units are now used in the generation of spectra. Also:

- Doc strings have been cleaned up.
- `planck` has been "fixed" to produce spectral luminosity density rather than spectral radiance density. In reality this never mattered because the output was always normalised but meant the units were wrong (and would give an incorrect answer if called separately).
- Normalisation is now done consistently using `Sed` methods rather than `scipy`.

I've also updated the ruff workflow version to use the latest hash and fixed an issue introduced in #727 

## Issue Type

- Bug
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/emission_models/dust/emission.py
+++ b/src/synthesizer/emission_models/dust/emission.py
@@ -148,11 +148,7 @@ class EmissionBase:
         # multidiensional arrays properly)
         if bolometric_luminosity.value.ndim == 0:
             lnu = (
-                bolometric_luminosity.to("erg/s").value
-                * self._get_spectra(lam)._lnu
-                * erg
-                / s
-                / Hz
+                bolometric_luminosity.to("erg/s") * self._get_spectra(lam).lnu
             )
         else:
             lnu = (
@@ -164,6 +160,8 @@ class EmissionBase:
                 / s
                 / Hz
             )
+
+        print(bolometric_luminosity)
 
         # Create new Sed object containing dust emission spectra
         return Sed(lam, lnu=lnu)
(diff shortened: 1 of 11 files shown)


WillJRoper:
And the untis pay off: the `Greybody` had a rogue extra nu in it. The emissivity can be a function of `nu` but there is not `nu*emissivity`.

src/synthesizer/emission_models/dust/emission.py
```diff
@@ -358,40 +377,44 @@ def _lnu(self, nu: unyt_array) -> unyt_array:
                 The unnormalised spectral luminosity density.
 
         """
+        # Ensure we have units on nu
+        if not has_units(nu):
+            raise exceptions.InconsistentArguments(
+                f"Frequency must be a given with units. (type(nu)={type(nu)})"
+            )
 
-        return nu**self.emissivity * planck(nu, self.temperature)
```
aswinpvijayan, line 362: This is actually correct. That is the functional form. The units are supposedly corrected at normalisation. One thing we could do make it (nu/nu_0)**emissivity to make it unitless?

aswinpvijayan: See for the different parameterisation definitions of the IR emission in: https://ui.adsabs.harvard.edu/abs/2012MNRAS.425.3094C/abstract

WillJRoper: Sure, we could do that. The pure nu multiplication isn't dimensionally consistent so we should avoid it. Seems a weird parametrisation though. Can't look at the paper right now but I'll check it out.

aswinpvijayan: It is defined as a proportionality in their definition, so probably they did not care about the units.

WillJRoper: That'd do it 😂. We can just arbitrarily divide by Hz to maintain the previous behaviour?

stephenmwilkins (approved):

PR #727, opened 2024-09-10
BUGFIX: Fixing `Grid.get_spectra` edge case 

WillJRoper:
@stephenmwilkins found a bug where the result of `measure_bolometric_luminosity` differed from the numpy result but only when the `Sed` was sourced from `Grid.get_spectra` which did not previously associate units to the grid point it made the `Sed` from. It is unclear exactly what causes this but passing units (which would be best practice anyway) fixes this issue immediately.

## Issue Type

- Bug

## Checklist

### DIFF
--- a/src/synthesizer/emission_models/operations.py
+++ b/src/synthesizer/emission_models/operations.py
@@ -591,9 +591,9 @@ class DustAttenuation:
         """
         # Unpack the tau_v value unpacking any attributes we need
         # to extract from the emitter
-        tau_v = 1
+        tau_v = 0
         for tv in this_model.tau_v:
-            tau_v *= getattr(emitter, tv) if isinstance(tv, str) else tv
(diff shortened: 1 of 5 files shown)


src/synthesizer/emission_models/operations.py
stephenmwilkins, line 1: Can you add a quick comment justifying these changes?

stephenmwilkins (changes requested):
Just add a comment about the changes to the dust.

src/synthesizer/emission_models/operations.py
WillJRoper: This fixes the multiplication which should have been an addition. Luckily everyone has been using individual models for different optical depths instead of the tuple of different optical depths. 

No idea how it survived this long 😅

stephenmwilkins (approved):

PR #733, opened 2024-09-13
Update abundances with plural metallicities axis

sophie-newman:
Here I have updated Abundances by replacing `metallicity` with `metallicities`. This allows `create_cloudy_input_grid` in `synthesizer-grids` to work.

## Issue Type
- Bug
- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/abundances/reference_abundance_patterns.py
+++ b/src/synthesizer/abundances/reference_abundance_patterns.py
@@ -25,7 +25,7 @@ class Asplund2009:
         self.bibcode = "2009ARA&A..47..481A"
 
         # total metallicity
-        self.metallicity = 0.0134
+        self.metallicities = 0.0134
 
         # logarthmic abundances, i.e. log10(N_element/N_H)
         self.abundance = {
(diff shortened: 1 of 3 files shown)


src/synthesizer/abundances/reference_abundance_patterns.py
```diff
@@ -25,7 +25,7 @@ def __init__(self):
         self.bibcode = "2009ARA&A..47..481A"
 
         # total metallicity
-        self.metallicity = 0.0134
+        self.metallicities = 0.0134
```
WillJRoper, line 28: This is where things get weird... since these are "total" metallicity it makes sense for it to be singular 😅. Are these always single numbers? 

What was the issue on the grids side? Does it need to automatically use the plural naming?

sophie-newman: Yeah it does look a bit odd doesn't it.

This is the error I was getting on the grids side:
`Traceback (most recent call last):
  File "/home/sophienewman/code/synthesizer-grids/src/synthesizer_grids/cloudy/create_cloudy_input_grid.py", line 230, in <module>
    abundances = Abundances(
                 ^^^^^^^^^^^
TypeError: Abundances.__init__() got an unexpected keyword argument 'metallicities'`

so we need to get Abundances to be okay with metallicities being the name of an axis

PR #737, opened 2024-09-24
Introduce a decorator for handling expected units

WillJRoper:
This PR introduces an `accepts` decorator. This decorator is modelled after `unyt.accepts` but introduces some Synthesizer specific functionality.

When using `accepts` you can define the units expected for an argument. The decorator will then first check units have been passed for that argument, if so it will perform a conversion to the expected unit (if not already in the that unit). 

This reduces boilerplate where we already check that units have been passed, perform conversions, and check units are compatible, and will also improve safety if employed across the codebase. 

There is no obvious performance penalty to this since we were already performing the wrapped checks in many places.

One limitation I've already spotted is arguments that are tuples/lists of unyt_quantities can't currently be handled in its current form. Should they be? I can envision some ways to make it work but it will introduce lots of `if isinstance` logic.

## Issue Type

- Enhancement

## Checklist


stephenmwilkins:
Can you point to a specific example of taking a tuple of values? Can these be changed to split into individual arguments or arrays?

WillJRoper:
For instance, `window` [here](https://github.com/flaresimulations/synthesizer/blob/main/src/synthesizer/sed.py#L530) but it's far from the only example. I could make it work by passing a tuple to `accepts` which would mean it knows how to interpret it. 

Splitting it out into separate arguments would work in that example but then you're tying yourself to not having tuples. Arrays could work but then you have mutable arguments for something that should immutable and that feels icky.

WillJRoper:
Tuples and list arguments are now handled properly.

christopherlovell (approved):
Looks good, haven't checked every doc or example explicitly but all tests are passing

PR #741, opened 2024-09-25
Added luminosity function integration test 

christopherlovell:
Adds an integration test that produces the g-band luminosity function for an entire CAMELS box. Will not run automatically -users must download the data and run manually. This is intended as an integration test  for when e.g. photometry pipelines from particle data show strange behaviour. It contains the LF produced from the last commit, with hash key provided. 

Additionally, I have updated the camels download scripts and example data to use a CV box, rather than LH, so we can better compare with existing results. 

## Issue Type
- Enhancement

## Checklist


WillJRoper (approved):
This looks good to me. Just need to wrangle that failing example.

More generally I do want to remove the CAMELS data from the repo. We could remove it entirely from the history... but I'm not 100% sure I want to deal with that 😂 if we remove it I can add `*.hdf5` to the gitignore and we'll never have to worry about it again!

PR #742, opened 2024-09-26
Add bolometric luminosity as property

stephenmwilkins:
Adds `bolometric_luminosity` as a property method deprecating `measure_bolometric_luminosity`. This is less efficient since the quantity will always be calculated but, as we've seen, this is safer. 

I'm somewhat ambivalent about whether `measure_bolometric_luminosity` should be deprecated since it allows the bolometric luminosity to be calculated using different methods. However, we're probably never going to use anything other than trapezoid?!

TODO: find all other instances of `measure_bolometric_luminosity` and replace.

## Issue Type
- [x] Enhancement

## Checklist

### DIFF
--- a/docs/source/sed/sed_example.ipynb
+++ b/docs/source/sed/sed_example.ipynb
@@ -197,7 +197,7 @@
     "\n",
     "There are a number of helper methods on ``Sed`` for calculating common derived properties. We provide some examples below.\n",
     "\n",
-    "We can calculate the integrated bolometric luminosity of the Sed using the `bolometic_luminosity` property methof:"
+    "We can calculate the integrated bolometric luminosity of the Sed using the `bolometric_luminosity` property method:"
    ]
   },
   {
--- a/docs/source/sed/sed_example.ipynb
+++ b/docs/source/sed/sed_example.ipynb
@@ -197,7 +197,7 @@
     "\n",
     "There are a number of helper methods on ``Sed`` for calculating common derived properties. We provide some examples below.\n",
     "\n",
-    "We can calculate the integrated bolometric luminosity of the Sed:"
+    "We can calculate the integrated bolometric luminosity of the Sed using the `bolometic_luminosity` property methof:"
    ]
   },
   {
@@ -207,7 +207,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sed.measure_bolometric_luminosity()"
+    "sed.bolometric_luminosity"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


docs/source/sed/sed_example.ipynb
```diff
@@ -197,7 +197,7 @@
     "\n",
     "There are a number of helper methods on ``Sed`` for calculating common derived properties. We provide some examples below.\n",
     "\n",
-    "We can calculate the integrated bolometric luminosity of the Sed:"
+    "We can calculate the integrated bolometric luminosity of the Sed using the `bolometic_luminosity` property methof:"
```
WillJRoper, line 200: Typo

src/synthesizer/sed.py
```diff
@@ -439,6 +435,33 @@ def shape(self):
...
+        # array.
+        integral = -integrate_last_axis(
+            self._nu,
+            self._lnu,
+            nthreads=1,
+            method="trapz",
+        )
+
+        # return the bolometric luminosity with units
+        return integral * self.lnu.units * self.nu.units
```
WillJRoper, line 454: Here is where you could instantiate a `Units` object and extract the bolometric luminosity units for conversion. However, it would be weird for the user to set a system where the combination of `lnu` and `nu` units didn't give the bolometric luminosity 😅 I'm tempted to say we leave it like this?

src/synthesizer/sed.py
```diff
@@ -477,6 +500,11 @@ def get_lnu_at_lam(self, lam, kind=False):
         """
         return interp1d(self._lam, self._lnu, kind=kind)(lam) * self.lnu.units
 
+    @deprecated(
+        message=(
+            "Deprecated in favour of bolometric_luminosity" " propery method"
```
WillJRoper, line 505: One string if it fits on one line.

WillJRoper (changes requested):

docs/source/sed/sed_example.ipynb
stephenmwilkins: Done

src/synthesizer/sed.py
stephenmwilkins: Done

src/synthesizer/sed.py
stephenmwilkins: Do you mean e.g.:

```
from synthesizer.units import Units
units = Units()
units.luminosity
```
or is there a faster way of extracting the units?

Incidentally, in an ideal world I think we would have just defined the units of `luminosity` and then aliased `bolometric_luminosity` to it.

WillJRoper: That would be the way to do it. However, I've wanting for a while now to categorise the unit system so that the user defines `spatial`, `luminosity`, `wavelength` etc. and then those get assigned to the specific attribute within `Units` (or aliased as you say).

If I did that then the code you have here is already sufficient.

WillJRoper (approved):

PR #750, opened 2024-10-07
Integrating imaging into EmissionModel framework (+abstract base class)

WillJRoper:
This PR integrates `EmissionModels` into the imaging functionality. 

Previously (and as demonstrated in the LRD pipeline), the imaging hadn't caught up to the rest of the package leading to label mismatches between photometry/spectra and the images. With this PR the imaging now ingests an `EmissionModel` and produces images following the same rules all other emissions do. 

Since the imaging can be expensive, unlike the spectra, imaging has a `limit_to` argument that can take a single model label and only produce images for that spectra type.

In addition to the imaging functionality, this PR consolidates a lot of duplicated functionality across different components into a `Component` abstract base class. This was spurred on by the fact the imaging methods would have had to be introduced in multiple places but now only needs wrappers on a `BaseGalaxy` and a `Component` (along with all other common attributes and methods. 

## Issue Type

- Enhancement

## Checklist


src/synthesizer/base_galaxy.py
```diff
@@ -967,6 +971,224 @@ def get_lines(
...
+    def get_images_luminosity(
+        self,
+        resolution,
+        fov,
+        emission_model,
+        img_type="smoothed",
+        kernel=None,
+        kernel_threshold=1,
+        nthreads=1,
+        limit_to=None,
```
aswinpvijayan, line 983: missing docstring for that

aswinpvijayan (changes requested):

aswinpvijayan (approved):

PR #761, opened 2024-10-16
Generalise ionising photon calculation to arrays

stephenmwilkins:
At the moment Sed.calculate_ionising_photon_luminosity() doesn't work for arrays of spectra. This fixes this.

## Issue Type
- [x] Bug

## Checklist


src/synthesizer/sed.py
```diff
@@ -1314,18 +1314,25 @@ def calculate_ionising_photon_production_rate(
...
-            ionisation_wavelength.to(angstrom).value, x, y
-        )
-
         # Restrict arrays to ionisation regime
         x = x[ionisation_mask]
-        y = y[ionisation_mask]
+        if len(y.shape) == 1:
+            y = y[ionisation_mask]
+        else:
+            y = y[:, ionisation_mask]
```
WillJRoper, line 1322: This only works for 1D and 2D now, what about 3D?  

Use
```
masked_array = arr[..., mask]
```
to mask the final axis of an arbitrarily dimensional array.

src/synthesizer/sed.py
```diff
@@ -1314,18 +1314,25 @@ def calculate_ionising_photon_production_rate(
...
+
+        # Add a final data point at the ionising energy to ensure full
+        # coverage.
+        x0 = ionisation_wavelength.to(angstrom).value
+        if len(y.shape) == 1:
```
WillJRoper, line 1327: Same with this block, it'll fail for 3D arrays. You could use scipy and `interp1d` with the axis argument or I think you could just use pure numpy with `apply_along_axis`
```
y0 = np.apply_along_axis(lambda y_: np.interp(x0, x, y_), axis=-1, arr=y)
``` 
https://numpy.org/doc/2.0/reference/generated/numpy.apply_along_axis.html

WillJRoper (changes requested):

src/synthesizer/sed.py
stephenmwilkins: Fixed.

src/synthesizer/sed.py
stephenmwilkins: Fixed. See comment below.

stephenmwilkins:
I've generalised it nD and added some examples to the docs demonstrating this. However, this revealed that several other methods on Sed don't work in 3D. I'll raise an issue for these.

docs/source/sed/sed_example.ipynb
```diff
@@ -509,13 +509,47 @@
...
+   "execution_count": null,
+   "id": "0cbf660f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sed3 = grid.get_sed(spectra_type='incident')\n",
+    "print(sed3.lnu.shape)\n",
+    "\n",
+    "print(sed3.measure_window_lnu((1400, 1600) * Angstrom))\n",
+    "# print(sed3.measure_beta())  # DOES NOT WORK\n",
```
WillJRoper, line 544: Can't merge stuff like this into main for it to appear in the docs. Please remove it or fix it.

WillJRoper (changes requested):

docs/source/sed/sed_example.ipynb
stephenmwilkins: Done

WillJRoper (approved):

PR #762, opened 2024-10-16
Scale BH spectra by the `bolometric_luminosity`

stephenmwilkins:
At present BH spectra - except in UnifiedAGN - are unscaled (scaled by unity) since previously they were the true luminosities predicted by the model. However, this was changed in favour of having grids normalised to unity. Consequently this PR now scales all BH by the `bolometric_luminosity`. 

This requires a bit more testing to check for impacts on UnifiedAGN and would be cleaner if previous PRs (which are merged here) are merged to main first.

## Issue Type
- [x] Bug

## Checklist


src/synthesizer/sed.py
```diff
@@ -1314,18 +1314,25 @@ def calculate_ionising_photon_production_rate(
         x = self._lam
```
WillJRoper, line 1314: Can't merge this because it includes #761

src/synthesizer/particle/blackholes.py
```diff
@@ -396,9 +396,9 @@ def _prepare_sed_args(
             for prop in props
         ]
 
-        # For black holes mass is a grid parameter but we still need to
-        # multiply by mass in the extensions so just multiply by 1
-        mass = np.ones(npart, dtype=np.float64)
+        # For black holes the grid Sed are normalised to 1.0 so we need to
+        # scale by the bolometric luminosity.
+        bol_lum = self.bolometric_luminosity.value
```
WillJRoper, line 401: You also need to do this in `parametric/blackhole.py` in the same functions at lines 198 and 335.

WillJRoper (changes requested):
Looks great. Just needs the changes mirrored in the parametric object and #761 to be removed. Just run:
```
git checkout main -- src/synthesizer/sed.py
```
to remove those changes.

src/synthesizer/particle/blackholes.py
stephenmwilkins: Done

WillJRoper:
This is ready but can't be merged because it still contains #761 (including the updates). You can just checkout the normal version of those files to remove these changes.

WillJRoper (approved):

PR #763, opened 2024-10-16
Introducing the `Pipeline` replacement

WillJRoper:
This PR introduces the `Pipeline` class, a class for automating pipelines to reduce the amount of boilerplate a user needs to implement every time they want to write a pipeline for a new simulation or model.

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/emission_models/base_model.py
+++ b/src/synthesizer/emission_models/base_model.py
@@ -1446,18 +1446,8 @@ class EmissionModel(Extraction, Generation, DustAttenuation, Combination):
             group (h5py.Group):
                 The group to save the model to.
         """
-        # First off call the operation to save operation specific attributes
-        # to the group
-        if self._is_extracting:
-            self.extract_to_hdf5(group)
-        elif self._is_combining:
-            self.combine_to_hdf5(group)
-        elif self._is_dust_attenuating:
-            self.attenuate_to_hdf5(group)
-        elif self._is_dust_emitting:
-            self.generate_to_hdf5(group)
-        elif self._is_generating:
-            self.generate_to_hdf5(group)
+        # First off write out the operation specific attributes
+        self._to_hdf5(group)
 
         # Save the model attributes
         group.attrs["label"] = self.label
--- a/src/synthesizer/emission_models/base_model.py
+++ b/src/synthesizer/emission_models/base_model.py
@@ -1446,8 +1446,18 @@ class EmissionModel(Extraction, Generation, DustAttenuation, Combination):
             group (h5py.Group):
                 The group to save the model to.
         """
-        # First off write out the operation specific attributes
-        self._to_hdf5(group)
+        # First off call the operation to save operation specific attributes
+        # to the group
+        if self._is_extracting:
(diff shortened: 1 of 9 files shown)


WillJRoper:
For an out in the wild example see: https://github.com/WillJRoper/flares-lrds/blob/main/flares_survey.py

WillJRoper:
This is actually ready now. Had missed some doc updates the first time I readied it.

WillJRoper:
There are some changes outside the ``Pipeline`` introduction itself that were needed to make it work efficiently. In summary these are:

- Passing nthreads around where it was missing.
- Enabling I/O for saving instances of classes to HDF5.
- Allowing you to select a subset of a container.

WillJRoper:
FYI, I want to add some options for limiting what is written out (probably flags passed to `write`) but I'll implement this in a future PR. The motivation being... with the images all saved my FLARES pipeline above is saving 7.2TB before doing z=5 😬

src/synthesizer/emission_models/base_model.py
```diff
@@ -1434,6 +1438,62 @@ def fix_parameters(self, **kwargs):
...
+        """
+        Save the model to an HDF5 group.
+
+        Args:
+            group (h5py.Group):
+                The group to save the model to.
+        """
+        # First off call the operation to save operation specific attributes
+        # to the group
+        if self._is_extracting:
```
christopherlovell, line 1451: I'm not sure why each of these models has to have a dedicated to_hdf5 method, can't an extraction / combination / etc. just have a general `to_hdf5` method, and then call it once here?

src/synthesizer/pipeline/pipeline.py
```diff
@@ -0,0 +1,1801 @@
...
+            for f, img in imgs.items():
+                self.images_flux_psf["BlackHole"][spec_type][f] = unyt_array(
+                    img
+                )
+
+        # Done!
+        self._got_images_flux_psf = True
+        self._took(start, "Applying PSFs to flux images")
+
+    def get_lnu_data_cubes(self):
```
christopherlovell, line 1407: Why the4 different naming convention here compared to the spectra and images?

christopherlovell (changes requested):

src/synthesizer/emission_models/base_model.py
WillJRoper: It reduces the boilerplate to have a targeted writer for each operation. Each operation only has certain attributes populated. If you didn't isolate these then you'd have to have a check for each attr you're writing.

src/synthesizer/pipeline/pipeline.py
WillJRoper: In short, no idea 😅 I had ignored them after adding the `NotImplemented` error.

src/synthesizer/emission_models/base_model.py
WillJRoper: Misunderstood, now fixed to use a single private `_to_hdf5` method on each operation

christopherlovell (approved):

PR #764, opened 2024-10-17
Fix for periodic boundary

aswinpvijayan:
## Issue Type
- Bug

The previous version when loading the particle data did not correct for the periodic boundary. So in a few galaxies not all particles were being counted.

Also calculates the dust-to-metal ratio in the load and adds it to the galaxy.

## Checklist

### DIFF
--- a/src/synthesizer/load_data/load_eagle.py
+++ b/src/synthesizer/load_data/load_eagle.py
@@ -96,6 +96,10 @@ def load_EAGLE(
     if grpno.dtype == object:
         return []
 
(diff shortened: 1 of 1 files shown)


WillJRoper (commented):
Are you sure you pushed everything? Seems to only include changes to the DTM?

aswinpvijayan:
Yep, absolutely right, forgot that I had done a stash a while ago. Added it now

WillJRoper (approved):

PR #783, opened 2024-11-15
Vel shift spectra marco

marcoleonardi97:
## Issue Type

- Bug
- Document
- Enhancement

## Checklist


marcoleonardi97 (commented):
there's some old commit from my fork, couldn't remove them, hopefully they don't mess anything up in the new files

marcoleonardi97:
Added flags from the front-end (component.get_spectra) to the stars generate_lnu and generate_particle_lnu. Won't work yet because we need to parse velocities as an argument in prepare_sed_args. 
Generate_lnu also uses compute_integrated_sed instead of compute_particle_sed, need to modify that too?

WillJRoper:
> Generate_lnu also uses compute_integrated_sed instead of compute_particle_sed, need to modify that too?

I was thinking about this on my drive this morning. The easiest way to handle this would actually be to call your spectra generation function from within either compute_integrated_sed or compute_particle_sed if the velocity flag is true. (rather than have the whole separate script). That way we don't need such extensive edits and you don't need to worry so much about the front end (only what you've already done). 

To be clear I mean having something like this in the compute_integrated_sed and compute_particle_sed functions:
```
if (with_vel_shift) {
   <call your function>
 } else {
   <call_the_old_function>
}
```

src/synthesizer/extensions/particle_spectra_with_shift.c
```diff
@@ -42,10 +42,10 @@ int find_nearest_bin(double lambda, double *grid_wavelengths, int nlam) {
  * @param spectra: The output array.
  */
 static void shifted_spectra_loop_cic_serial(struct grid *grid, struct particles *parts,
-                                    double *spectra) {
+                                    double *spectra, c) {
```
WillJRoper, line 45: Needs to be `const double c` because its a floating point number we want precision on and we know it won't ever be changed.

src/synthesizer/extensions/particle_spectra_with_shift.c
```diff
@@ -141,15 +141,16 @@ static void shifted_spectra_loop_cic_serial(struct grid *grid, struct particles
       for (int ilam = 0; ilam < nlam; ilam++) {
         double shifted_lambda = shifted_wavelengths[ilam];
       /*Find nearest wavelength bin*/
-        int ilam_shifted = find_nearest_bin(shifted_lambda, wavelength, nlam); // should never be > nlam hopefully
+        int ilam_shifted = find_nearest_bin(shifted_lambda, wavelength, nlam); 
+        if (ilam_shifted >= nlam - 1) {continue}; // jumps out of the loop if out of bounds
```
WillJRoper, line 145: Need to handle `< 0` case too

src/synthesizer/particle/stars.py
```diff
@@ -565,8 +566,11 @@ def generate_lnu(
         )
 
         # Get the integrated spectra in grid units (erg / s / Hz)
-        lnu_particle = compute_integrated_sed(*args)
-
+        if shift=False:
+            lnu_particle = compute_integrated_sed(*args) 
+        else:
+            lnu_particle = compute_integrated_sed(*args) # need to modify this too?
```
WillJRoper, line 572: Yea, if we did do it this way then we need to call `compute_particle_seds_shifted` here (well in reality we want an integrated one here but we'll cross that bridge).

marcoleonardi97:
modified the particle_spectra.c directly so we can choose which functions to call when, i just duplicated the ngp and cic wrappers and made the respective shifted versions. (Not sure if having 2 functions for task is more of less efficient than having 1 with a flag, but should work anyways). 

added flag to generate_particle_lnu for black holes, 

only thing left to do (i think) should be reworking compute_integrated_sed in the same way of compute_particle_sed, then see how these all work when they are called in a galaxy object

WillJRoper (approved):

PR #789, opened 2024-11-21
Fixing lyman alpha escape fraction application

WillJRoper:
During the switch over to `EmissionModels` the lyman alpha escape fraction wasn't ported correctly and isn't being applied properly. This PR fixes this by introducing a `LymanAlphaEmission` model.

## Issue Type

- Bug

## Checklist


src/synthesizer/emission_models/__init__.py
```diff
@@ -1,51 +1,49 @@
+from synthesizer.emission_models.agn import AGN_MODELS
```
WillJRoper, line 1: Sorry, new `ruff` means this all got formatted. Only changes are the renaming of `LineContinuumEmission` to `LineContributionEmission`.

src/synthesizer/emission_models/stellar/models.py
```diff
@@ -78,24 +80,42 @@ class LineContinuumEmission(StellarEmissionModel):
...
 
         Args:
             grid (synthesizer.grid.Grid): The grid object to extract from.
+            fesc_ly_alpha (float): The escape fraction of Lyman-alpha.
             label (str): The label for this emission model.
             fesc (float): The escape fraction of the emission.
         """
+        print(fesc, fesc_ly_alpha)
+        # Create the child models
+        linecont_no_lyman = StellarEmissionModel(
```
WillJRoper, line 99: This the important bit, automatically defining the 2 contributions with their unique escape fractions.

src/synthesizer/grid.py
```diff
@@ -306,16 +306,46 @@ def _get_spectra_grid(self, spectra_to_read):
...
             self.available_spectra.append("total")
 
+            # The nebular continuum is the nebular emission with the line
+            # contribution removed
             self.spectra["nebular_continuum"] = (
                 self.spectra["nebular"] - self.spectra["linecont"]
             )
             self.available_spectra.append("nebular_continuum")
 
+            # Get the lyman alpha wavelength elements and create a mask for
```
WillJRoper, line 323: Also here, I've now made a lyman free version of `"linecont"`

christopherlovell:
I've updated things to clarify some of the confusion. Aside from in the grid (where the naming is set by cloudy) I've removed all references to `linecont` and replaced these with `nebular_line`. The methods have similarly been renamed, to `NebularLineEmission`.

I have also fixed the handling of lyman alpha escape fractions, ensuring that the Nebular Line contribution at this wavelength is properly reduced, and combined with the full line emission spectra. I have also accounted for when `fesc` is non-zero, and how this is handled when `fes_ly_alpha` is also non-zero, properly reducing the overall line emission, and then applying the extra lyman alpha escape fraction.

src/synthesizer/emission_models/stellar/models.py
```diff
@@ -78,26 +80,71 @@ class LineContinuumEmission(StellarEmissionModel):
...
         """
-        StellarEmissionModel.__init__(
-            self,
-            grid=grid,
-            label=label,
-            extract="linecont",
-            fesc=fesc,
-            **kwargs,
-        )
+        if fesc_ly_alpha < 1.0:
```
WillJRoper, line 97: The way this has always been used previously had this logic in the `NebularEmission`, has this changed? Is there a situation when you would be using this model without `fesc_ly_alpha=1`? If not we can simplify things.

src/synthesizer/emission_models/stellar/models.py
```diff
@@ -78,26 +80,71 @@ class LineContinuumEmission(StellarEmissionModel):
...
+            # Get the lyman alpha wavelength elements and create a mask for
+            # the line (to account for spreading between bins)
```
WillJRoper, line 99: Comment doesn't make sense (maybe move the spreading bracket to the branches below?

src/synthesizer/emission_models/base_model.py
```diff
@@ -414,10 +421,12 @@ def _init_operations(
                 luminosity when computing dust emission.
             fesc (float):
                 The escape fraction.
+            lam_mask (ndarray):
+                The mask to apply to the wavelength array.
         """
         # Which operation are we doing?
         if self._is_extracting:
-            Extraction.__init__(self, grid, extract, fesc)
+            Extraction.__init__(self, grid, extract, fesc, lam_mask)
```
WillJRoper, line 429: We probably want to propagate the lam mask to all 4 operations to give full flexibility (e.g. applying attenuation in specific ranges). Probably niche but it's maybe a bit confusing conceptually that you can pass a lam_mask into any model but it only gets used from an extraction.

src/synthesizer/emission_models/operations.py
```diff
@@ -53,6 +55,9 @@ def __init__(self, grid, extract, fesc):
         # Attach the escape fraction
         self._fesc = fesc
 
+        # Attach the wavelength mask
+        self._lam_mask = lam_mask
```
WillJRoper, line 59: Following from comment above, probably better to attach this at the base model if we do allow all operations to use a wavelength mask.

src/synthesizer/emission_models/stellar/models.py
```diff
@@ -260,41 +307,43 @@ def __init__(
...
+        if nebular_continuum is None:
+            nebular_continuum = NebularContinuumEmission(
+                grid=grid,
+                fesc=fesc,
+                **kwargs,
+            )
+
+        # If we have a Lyman-alpha escape fraction then calculate the
+        # updated line emission and combine with the nebular continuum.
+        if fesc_ly_alpha < 1.0:
```
WillJRoper, line 336: Alternative to the above: Nebular is always a combination of the lines and continuum and we remove this logic here?

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -113,7 +114,7 @@ def __init__(
         self,
         grid,
         tau_v,
-        dust_curve,
+        dust_curve=PowerLaw(),
```
WillJRoper, line 117: Why force a default here? Surely better to make the user think about it? Especially because a `PowerLaw` also assumes a default slope.

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -853,7 +856,7 @@ def _make_nebular(self, **kwargs):
             mask_op="<",
             fesc=self._fesc,
             fesc_ly_alpha=self._fesc_ly_alpha,
-            linecont=young_line_cont,
+            linecont=young_neb_line,
```
WillJRoper, line 859: Incorrect arg name, this is the origin of the error in the docs I think. `linecont` -> `nebular_line`

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -865,7 +868,7 @@ def _make_nebular(self, **kwargs):
             mask_op=">=",
             fesc=self._fesc,
             fesc_ly_alpha=self._fesc_ly_alpha,
-            linecont=old_line_cont,
+            linecont=old_neb_line,
```
WillJRoper, line 871: Same here

src/synthesizer/grid.py
```diff
@@ -306,11 +306,15 @@ def _get_spectra_grid(self, spectra_to_read):
         # If a full cloudy grid is available calculate some
         # other spectra for convenience.
         if self.reprocessed:
+            # The total emission (ignoring any dust reprocessing) is just
```
WillJRoper, line 309: Do we actually need this? I don't know that we ever use it. Is it more of a "nice to see when plotting from the `Grid`" situation?

src/synthesizer/parametric/stars.py
```diff
@@ -502,6 +511,10 @@ def generate_lnu(
             axis=0,
         )
 
+        # Apply the wavelength mask if provided
+        if lam_mask is not None:
+            spectra[~lam_mask] = 0.0
```
WillJRoper, line 516: Could we not be a bit cleverer here and limit the grid spectra with the mask in `prepare_sed_args` that is fed to the extension. That way we never do the calculation on the masked elements, instead we just add them into a zeros array here to ensure the shape is consistent (like when we apply a particle mask).

src/synthesizer/particle/blackholes.py
```diff
@@ -672,6 +675,13 @@ def generate_particle_lnu(
         spec = np.zeros((self.nbh, masked_spec.shape[-1]))
         spec[mask] = masked_spec
 
+        # Apply the wavelength mask if provided
+        spec[:, ~lam_mask] = 0.0
+
+        # TODO: Wavelength masking could be done before the C function call to
```
WillJRoper, line 681: HA, like you say here 😂 but doesn't require any grid modifications. We can just apply the lam mask in the argument preparation method.

src/synthesizer/particle/stars.py
```diff
@@ -838,9 +841,17 @@ def generate_lnu(
...
 
         # Get the integrated spectra in grid units (erg / s / Hz)
-        lnu_particle = compute_integrated_sed(*args)
+        lnu = compute_integrated_sed(*args)
+
+        # Apply the wavelength mask if provided
+        if lam_mask is not None:
+            lnu[~lam_mask] = 0.0
+
+        # TODO: Wavelength masking could be done before the C function call to
```
WillJRoper, line 850: Same story

src/synthesizer/particle/stars.py
```diff
@@ -1424,6 +1434,13 @@ def generate_particle_lnu(
 
         toc("Masking spectra and adding contribution", start)
 
+        # Apply the wavelength mask if provided
+        spec[:, ~lam_mask] = 0.0
+
+        # TODO: Masking could be done before the C function call to reduce
```
WillJRoper, line 1440: And same again.

WillJRoper (commented):
Looking good. A few comments, they seem more thorough than they are. Most comments are repeating the same thing in different places so seems like a bunch

src/synthesizer/emission_models/stellar/models.py
christopherlovell: if you want the lines, you need to know what the lyman alpha escape fraction is so you can modify the luminosity of that line, so yes it needs to be here. Arguably it only needs to be here, since the nebular emission should derive from the combination of the lines and continuum

WillJRoper: Sounds good to me. Lets kill the logic in the `NebularEmission` model then. Just simplifies things a bit if we don't have multiple of the same branching logic.

src/synthesizer/emission_models/stellar/models.py
christopherlovell: The logic here was that just grabbing an array from the grid will be quicker, but I guess we're grabbing the other arrays already. 

As an aside, a nice addition to the test suite would be something that grabs these from the emission model and comparea to the `nebular` key from the grid.

src/synthesizer/parametric/stars.py
christopherlovell: See comments on all other instances :) feel free to implement

src/synthesizer/grid.py
christopherlovell: I think so yeah. I'd keep it

src/synthesizer/emission_models/stellar/pacman_model.py
christopherlovell: Ah, does it do the strong concatenation automatically?

src/synthesizer/emission_models/stellar/pacman_model.py
christopherlovell: I'm open to using a different default (Calzetti?), but I think we should provide one. 99% of users will not care.

WillJRoper: Fair, I can be convinced to use a default. I was working on a "make them choose" ethos through all the models 😂

src/synthesizer/emission_models/stellar/pacman_model.py
WillJRoper: Nah, it's that the keyword argument is checked for `nebular_line` so passing this to `linecont` will make the model try and make the same model twice. (I think anyway)

src/synthesizer/emission_models/base_model.py
christopherlovell: Kind of, but if you're masking by wavelength you probably want to do enforce that on an extraction or generation step anyway. Doing it on an attenuation seems a bit strange

WillJRoper: Yea, I resolved this because I came to the conclusion if it ever is needed for some reason it'll be trivial to add then.

WillJRoper:
This is good to go now @christopherlovell.

PR #792, opened 2024-11-25
Update SVO load to check units on transmission

### LINKED ISSUE #523: Some SVO Filters have transmission >1
Certain filter curves from SVO have transmission values in the 10s (e.g. ~60 for GALEX being an example).

I wonder if some transmission arrays are quoted in percent. If so we should probably have something like

```
if self.t.max() > 1:
    self.t /= 100
```

For SVO filters specifically.

christopherlovell:
Fixes #523 by ensuring that the loaded transmission doe snot have units. Required loading the XML directyl, rather than the ascii tables, to obtain the unit information.

## Issue Type
- Bug

## Checklist


WillJRoper (approved):
Phenomenal

PR #800, opened 2024-12-13
BUGFIX: Fixing the accepts decorator for list/tuple args

WillJRoper:
When `accepts` was used with a list arg it was being appended to instead of having values set. This is now fixed.

## Issue Type

- Bug

## Checklist

### DIFF
--- a/src/synthesizer/units.py
+++ b/src/synthesizer/units.py
@@ -533,15 +533,15 @@ def _check_arg(units, name, value):
         IncorrectUnits
             If the argument has incompatible units.
     """
+    # Early exit if the argument isn't in the units dictionary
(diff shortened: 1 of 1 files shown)


src/synthesizer/units.py
```diff
@@ -533,16 +533,16 @@ def _check_arg(units, name, value):
         IncorrectUnits
             If the argument has incompatible units.
     """
+    # Early exit if the argument isn't in the units dictionary
```
WillJRoper, line 536: Moved this to exit one branch quicker when checking non-unyt args.

aswinpvijayan (approved):

PR #804, opened 2024-12-19
Introducing velocity shifts in particle spectra (Will's modified version)

### LINKED ISSUE #364: Add broadening / velocity offset capability
We will ultimately want to add dynamical and thermal line broadening. 

I'm not sure where is best to add this but I think this might be best as a pair of methods on `Sed` which can be called from a `Galaxy`. I think there are two things we would like to implement in the first instance:

- **Thermal broadening** this can be implemented by simply convolving the spectrum by a gaussian with a given velocity. However, this should only apply to the nebular component.
- **Dynamical broadening** we should be able to shift the SED by a line-of-sight velocity offset (e.g. for a star particle moving relative to the central potential). For an array of SEDs we need to be able to provide an array of line-of-sight velocities.

For a particle galaxy the dynamical broadening could (or should) be added automatically if velocities are provided. 

For the thermal broadening we would probably want a method on `base_galaxy` that applies the broadening to the nebular component and recalculates the total spectrum.

WillJRoper:
This PR introduces velocity shifting of particle-based spectra. This is done during extraction where the velocity of each particle is used to shift the spectra. 

Some important points:

- The line of sight along which the velocity is taken is the z-axis to match all other line of sight calculations. Any other orientations can be achieved using the rotation methods. 
- As it stands the velocity shifting is turned on in a call to `get_spectra`. I'm undecided whether this should also be a flag on an emission model. I'm open to arguments either way and possibly adding this in the future. It is turned off by default.
- If an integrated spectrum is generated with velocity shift turned on then behind the scenes a particle-based spectra will first be generated, this is then summed to get the integrated spectra. This is unavoidable since each particle's spectra must be generated to account for its shift, meaning the usual weight grid calculation can't be used. We could in the future add an integrated velocity shift function which still calculates each particle's spectra but adds it to a single spectra array removing the memory overhead of creating each one but I see this as a low priority for now. 
- Tests have been added that show the flux is conserved. However, flux is not conserved at the very edges of the array where flux can be shifted outside the wavelength range. 
- Examples have been added to show the difference with and without the velocity shift and how rotating the galaxy effects it.

Closes #364 

## Issue Type

- Enhancement

## Checklist


WillJRoper:
@marcoleonardi97 I've had a play today. In doing so I split off into a different branch because I didn't want to edit your work but instead make the modifications separately so you can compare. 

There wasn't actually much at all holding it back from compiling. A few syntax errors but once those were fixed it was clear sailing to compile. 

Here I have:
- Done some small housekeeping. 
- Split off the mapping of shifted indices. This is done so we don't do it for each grid cell in the CIC versions. If we did it for every grid cell we'd repeat the same calculation `2^ndim` times. A big subtlety.
- The grid wavelengths needed passing so those have now been passed through to the C. 
- Swapped your `find_nearest_bin` for a binary search which will be faster. In big O notation, your loop version was at worst `O(N)` if the element you're looking for is at the end. The binary search is `O(log(n)` so much more efficient. (the binary search returns the upper bin so that required some minor modifications too).
- I added an example just to prove it works. 

By all means, move to using this branch or use it as a template, I don't mind. We can probably simplify the code a little by combining stuff together. I'm sure there's some time to be saved, so feel free to look at the performance.  

I need to fix some unrelated stuff that is now failing because of these modifications (mostly missing arguments in the integrated spectra stuff). Don't worry about that! 

Here's some plots showing your code working:

SCR-20241219-qlwn

Zoomed right in on a line

SCR-20241219-qlcd

WillJRoper:
Thanks for all the hard work @marcoleonardi97. I've now got this all sorted and passing. It'll now be reviewed by another member of the team before being merged into the main branch.

marcoleonardi97:
@WillJRoper thanks for cleaning it up for me ! can't wait to get some actual results

christopherlovell:
For the record, I think this should be a flag on an emission model, but happy to get this in for now

src/synthesizer/extensions/particle_spectra.c
```diff
@@ -20,6 +20,156 @@
...
 #include "weights.h"
 
+/**
+ * @brief Find nearest wavelength bin for a given lambda, in a given wavelength
+ * array. Used by the spectra loop functions when considering doppler shift
+ *
+ * Note: binary search returns the index of the upper bin of those that straddle
+ * the given lambda.
+ */
+int find_nearest_bin(double lambda, double *grid_wavelengths, int nlam) {
```
christopherlovell, line 30: Not essential, but this could be named better

src/synthesizer/extensions/particle_spectra.c
```diff
@@ -432,6 +893,135 @@ static void spectra_loop_ngp_omp(struct grid *grid, struct particles *parts,
 }
 #endif
 
+/**
+ * @brief This calculates particle spectra using a nearest grid point approach.
+ * This is the version of the function that accounts for doppler shift.
+ * This is the serial version of the function.
```
christopherlovell, line 899: should be the parallel version?

src/synthesizer/extensions/particle_spectra.c
```diff
@@ -565,10 +1202,119 @@ PyObject *compute_particle_seds(PyObject *self, PyObject *args) {
...
+  PyObject *grid_tuple, *part_tuple;
+  PyObject *py_vel_shift;
+  PyObject *py_c;
+  PyArrayObject *np_grid_spectra, *np_lam;
+  PyArrayObject *np_fesc;
+  PyArrayObject *np_velocities;
+  PyArrayObject *np_part_mass, *np_ndims;
+  char *method;
+
+  if (!PyArg_ParseTuple(args, "OOOOOOOOiiisiO", &np_grid_spectra, &np_lam,
```
christopherlovell, line 1242: What is this line doing?

examples/particle/plot_velocity_broadened_spectra.py
```diff
@@ -0,0 +1,106 @@
+"""
+Plot velocity broadened spectra
+===============================
+
+This example shows how to compute line of sight dust surface densities,
```
christopherlovell, line 5: Incorrect doctrsing

christopherlovell (changes requested):

src/synthesizer/extensions/particle_spectra.c
WillJRoper: This is how you read Python arguments into C, the string tells it what sort of Python object to expect and then every other argument is the variable to populate.

src/synthesizer/extensions/particle_spectra.c
WillJRoper: Done

src/synthesizer/extensions/particle_spectra.c
WillJRoper: Done

examples/particle/plot_velocity_broadened_spectra.py
WillJRoper: Done

WillJRoper:
Agreed on the emission model having a flag. That's easy enough to do (and then the argument currently being used becomes an override like the others). I can do that separately very easily.

christopherlovell (approved):

PR #815, opened 2025-01-28
Added the ability to blend lines based on wavelength bins

### LINKED ISSUE #561: Add method to LineCollection to blend lines
Often emission lines are blended with each other. We could easily mimic this by adding a method on `LineCollection` which creates a new `LineCollection` with blended lines. The free argument would be the wavelength resolution expressed in some way.

WillJRoper:
Added a method on `LineCollection` to bin lines and then blend any lines in the same bin together.

While implementing this I brought the lines module in line with everywhere else in terms of `__str__` methods. In doing so I found some hidden edge cases which have also been fixed.

Closes #561 

## Issue Type

- Enhancement

## Checklist


stephenmwilkins (approved):
Looks good, do I understand correctly that the combined line ID is the concatenation of the individual line IDs?

WillJRoper:
>Looks good, do I understand correctly that the combined line ID is the concatenation of the individual line IDs? 
Yup, that's always been the case for >=doublets.

PR #821, opened 2025-02-03
Introduces abundances dictionary as attribute on `Gas` and `Stars`

### LINKED ISSUE #603: Allow stars and gas particles to hold detailed chemical abundances
I would suggest adding a `metals` or `abundances` attribute which is a dictionary containing the abundance of an arbitrary set of elements.

WillJRoper:
Allowing for future functionality around complex abundances in baryons. 

- Both parametric and particle `Stars` can now carry an abundances dictionary.
- `s_oxygen` and `s_hydrogen` are now deprecated and will temporarily update the `abundances` dict. 
- Particle `Gas` can also carry the abundances. Note there is no real parametric `Gas` implementation for now but this is easily ported when there is by following the component implementation. 
- Closes #603 

## Issue Type

- Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -29,6 +29,11 @@ class StarsComponent(Component):
         metallicities (Quantity)
             The metallicity of each stellar particle/bin
             (particle/parametric).
+        abundances (dict)
+            A dictionary of the abundances of each element in the stars.
+            The keys are the element symbols and the values are the
+            abundances of each element in the stars (0-1 where 1 is 100%
+            of the stars is composed of that element).
     """
 
     # Define quantities
@@ -40,6 +45,7 @@ class StarsComponent(Component):
         ages,
         metallicities,
         _star_type,
+        abundances,
         **kwargs,
     ):
         """
@@ -51,6 +57,15 @@ class StarsComponent(Component):
             metallicities (array-like, float)
                 The metallicity of each stellar particle/bin
                 (particle/parametric).
+            _star_type (str)
+                The type of stars object (parametric or particle).
+                This is useful for determining the type of stars object
+                without relying on isinstance and possible circular imports.
+            abundances (dict)
+                The abundances of each element in the stars.
+                The keys are the element symbols and the values are the
+                abundances of each element in the stars (0-1 where 1 is 100%
+                of the stars is composed of that element).
         """
         # Initialise the parent class
         Component.__init__(self, "Stars", **kwargs)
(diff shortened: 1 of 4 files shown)


stephenmwilkins:
I wonder... if `abundances` were comprehensive it would be good to have it checked against metallicity. In fact, it could be used to calculate metallicity. Arguably that's exactly what we should do - when we have access to it.

stephenmwilkins:
I *think* abundance is often narrowly used to refer to abundances relative to H. It might be safer to call this something else, for example `mass_fraction(s)` to avoid confusion.

It gets worse since abundances are often quoted as `log10(X/H)+12`.

stephenmwilkins:
> I _think_ abundance is often narrowly used to refer to abundances relative to H. It might be safer to call this something else, for example `mass_fraction(s)` to avoid confusion.
> 
> It gets worse since abundances are often quoted as `log10(X/H)+12`.

The best approach is probably to have both (`abundances` `mass_fractions`), with one calculated from the other.

stephenmwilkins:
What does GADGET / SWIFT call these quantities?

WillJRoper:
> I wonder... if `abundances` were comprehensive it would be good to have it checked against metallicity. In fact, it could be used to calculate metallicity. Arguably that's exactly what we should do - when we have access to it.

I don't think we can ever bank on that. Who do you check for "comprehensive"? There's some stuff that is just down to the user

WillJRoper:
> A worry though, if I have a grid with metallicity and oxygen abundance as axes, will it still be possible to use a single abundance? or does the axis need to be a named attribute on the component?

By definition it has to be the latter.

WillJRoper:
I don't see the point in adding both attributes. The user can associate any arbitrary attribute to the component which can be associated to any Grid axis if they like. 

This was literally just to enable the user to carry abundances around in an exist dictionary if they want. It's not used anywhere. If its going to get this complex I'd argue we should nuke these arguments entirely and let the user set and use them freely. 

P.S. please use the review system instead of just commenting.

stephenmwilkins:
> I don't see the point in adding both attributes. The user can associate any arbitrary attribute to the component which can be associated to any Grid axis if they like.
> 
> This was literally just to enable the user to carry abundances around in an exist dictionary if they want. It's not used anywhere. If its going to get this complex I'd argue we should nuke these arguments entirely and let the user set and use them freely.
> 
> P.S. please use the review system instead of just commenting.

OK, I see. But doesn't that risk confusing the user? If they associate the component with an abundance / mass_fraction and use a grid using that information that might expect it to be used.

However, I suppose in that case synthesizer would throw a warning to complain that the grid axis is missing as an attribute?

WillJRoper:
>The check is simply whether it includes all the important elements.
That is not an well defined check 😅

WillJRoper:
I'm going to close this PR and close the issue itself. If you don't agree we can revisit at a later time.

PR #826, opened 2025-02-05
Simplifies the unit system by introducing "unit categories" 

### LINKED ISSUE #825: Unit system has become unwieldy
As the package has grown, so has the unit system. It is now pretty unwieldy. I've mentioned (a long time ago) adding an abstraction where we have the units listed by "category" (i.e. wavelength, spatial, frequency, mass etc.) and then use these entries to set units for each `Quantity` in the unit system. This would make it simpler to add new `Quantities` and also make it simpler for the user to modify the unit system without changing each individual attribute they want to change. 

One worry was that you may want different units for different attributes in the same category. However, I can't see a single example of that in the existing module, nor can I think of one we might introduce. Nonetheless, In the event that did become an issue, we can still keep the machinery which allows specific attributes to have their units overwritten.

I'd also like to add the default unit system as a yaml file. This could enable permanent altererations by the user via a command line interface. If they preferred a different standard unit (e.g. someone working with isolated galaxy sims would likely prefer kpc as the default spatial unit).

WillJRoper:
This PR switches the Unit system around slightly to improve maintainability and usability for anyone wanting modify the unit system. 

- Instead of defining the unit for each attribute we now define the unit for a "category" of attribute (see table below) 
- Instead of having to add any new Quantities to the class they are attached to, the `default_units` dict, the `Units` class, and the doc string of `Units` class, you now only have to define the `Quantity` on the class it is attached to. When you define a `Quantity` you now pass in the category. 
- A `Quantity` now holds its units from creation rather than accessing them on any get/set. This means that in the unlikely event someone goes out of their way to change the unit system mid analysis the old result will maintain their original unit.
- The default units system now carries a prettier print out.
- The unit system is now stored in a yaml file enabling a user to permanently modify the unit system if they wish.
- Closes #825 

```
+-------------------------------------------------+
|                  DEFAULT UNITS                  |
+-------------------------------+-----------------+
| Category                      | Unit            |
+-------------------------------+-----------------+
| spatial                       | Mpc             |
+-------------------------------+-----------------+
| mass                          | Msun            |
+-------------------------------+-----------------+
| time                   | yr              |
+-------------------------------+-----------------+
| velocity                      | km/s            |
+-------------------------------+-----------------+
| temperature                   | K               |
+-------------------------------+-----------------+
| angle                         | degree          |
+-------------------------------+-----------------+
| wavelength                    | Å               |
+-------------------------------+-----------------+
| frequency                     | Hz              |
+-------------------------------+-----------------+
| luminosity                    | erg/s           |
+-------------------------------+-----------------+
| luminosity_density_frequency  | erg/(Hz*s)      |
+-------------------------------+-----------------+
| luminosity_density_wavelength | erg/(s*Å)       |
+-------------------------------+-----------------+
| flux                          | erg/(cm**2*s)   |
+-------------------------------+-----------------+
| flux_density_frequency        | nJy             |
+-------------------------------+-----------------+
| flux_density_wavelength       | erg/(cm**2*s*Å) |
+-------------------------------+-----------------+
| mass_rate                     | Msun/yr         |
+-------------------------------+-----------------+
```
## Issue Type

- Enhancement

## Checklist


src/synthesizer/default_units.yml
```diff
@@ -0,0 +1,60 @@
...
+  # Spatial lengths and positions
+  spatial:
+    unit: Mpc
+
+  # Mass
+  mass:
+    unit: Msun
+
+  # Cosmic time (e.g. ages)
+  cosmic_time:
```
christopherlovell, line 11: Why `cosmic_time` and not just `time`?

WillJRoper: Just a distinction because seconds appear a lot in other units. I'm not wedded to the distinction though

christopherlovell: I would prefer just `time`. Not sure what is special about it being in a cosmic setting. And the fact that time appears in other compound units is... fine? we don't have cosmic lengths or cosmic masses...

WillJRoper: Yea, fair point. I'll make that change in a bit.

WillJRoper: Done now

christopherlovell (approved):

PR #827, opened 2025-02-07
Attenuation operation generalised to Transformation (+ butterfly effect)

### LINKED ISSUE #729: Some suggested improvements to emission_models
- Remove all additional potentially variable parameters e.g.  `covering_fraction_nlr`, `covering_fraction_blr`, `escape_fraction` from model initialisation.
- However, I think it would be useful to allow these to be fixed for an emission model. This may already work. This should over-
- For each model we should hard code descriptions of all the parameters are used specifically for the pre-made models including in the doc string. 
- More contentious, remove model-specific defaults from blackhole objects forcing a user to actually set them.

### LINKED ISSUE #818: Incorporate IGM attenuation into `EmissionModels`
Currently, you onlyhave to pass the IGM model to `get_fnu` (or `get_flux`) calls. I’m torn… because it really only makes sense to have IGM attenuation on observed fluxes, but at the same time, making it function inside a model makes a lot of sense in the framework we now have.

Note that the only thing we would need to do to make this change is add a premade model and make the IGM classes behave like the attenuation curves (i.e. have the same method structure as `PowerLaw` etc.)

WillJRoper:
In essence, this PR simply generalises the attenuation operation to encapsulate more general transformations. However, the knock-on effect of this was significant in both documentation and code. 

The number of change files is mostly due to the naming changes and documentation updates. Don't be fooled into thinking this is a huge change. The biggest real changes are within the ``emission_models`` module. 

- A `Transformer` abstract base class now exists to template out the attenuate, IGM etc transformers. This simply defines the common methods that are needed.
- This approach allowed for some operation simplifications but only if ``LineCollections`` had some missing methods added. This has been done and should now work from a single method for transformation reducing maintenance.
- This also includes a slight overhaul of the way `fixed_parameters` are treated following the points raised in #729 
- Parameters like `fesc` and `tau_v` are now considered to be primarily component attributes but they can be associated to models via the fixed parameter system (which can also take strings to find specific attributes for a specific model).
- `fesc` now has an inverse `escf` attribute on a component to enable easy escaped emission models that don't rely on model-level escape fractions (in fact this highlights that the escaped model has never really fully worked but no one has ever tested it using component-specific escape fractions). *This is more of a groundwork for the future change since we can use it to make escaped models more robust.*

This lays the groundwork for a more general simplification where each operation could be associated with an abstract base class defining what the "workhorse objects" must look like. The knock-on effect is better dependency injection and thus (counterintuitively) simpler code by adding extra classes. 

Closes #729 
Closes #818 

## Issue Type

- Enhancement

## Checklist


src/synthesizer/emission_models/base_model.py
```diff
@@ -918,7 +956,7 @@ def set_dust_props(
         Args:
             dust_curve (emission_models.attenuation.*):
                 A dust curve instance to apply.
-            apply_dust_to (EmissionModel):
+            apply_to (EmissionModel):
                 The model to apply the dust curve to.
             tau_v (float/ndarray/str/tuple):
```
aswinpvijayan, line 961: tau_v provided as argument in the docstring, but not in the function. Probably redundant here

WillJRoper: Good spot, killed the arg in the doc string

aswinpvijayan (approved):

src/synthesizer/components/component.py
```diff
@@ -39,11 +39,16 @@ class Component(ABC):
             A dictionary to hold the images in luminosity units.
         images_fnu (dict)
             A dictionary to hold the images in flux units
+        fesc (float)
+            The escape fraction of the component.
+        escf (float)
```
christopherlovell, line 44: Not sure about this naming convention. Also where is `escf` ever used?

christopherlovell (changes requested):

src/synthesizer/components/component.py
WillJRoper: Currently it isn't, my hope was to make the escaped model look for it by default. I should probably do that here.

Any other naming suggestion? I did think "inv_fesc" but it's not actually inverse.

christopherlovell (approved):

PR #829, opened 2025-02-10
run ruff

aswinpvijayan:
## Issue Type
The astropy dust extinction module is only for the wavelength range 1E2 - 1E8 Angstrom, which causes it to fail when the spectra goes over these ranges. Introduced some interpolation at shorter wavelengths and at longer wavelengths set it to the value at max wavelength (negligible impact at longer wavelength as the dust extinction almost -> 0).
- Enhancement

## Checklist


christopherlovell (approved):

WillJRoper:
This looks good. Is there a reason it's still in draft?

PR #835, opened 2025-02-12
Remove requirement for intrinsic to have non-zero escape fraction

christopherlovell:
This is a petty requirement. If you are using an emission model and want to test the dependence of your results on fesc, for example, then it will just fall over when it hits zero. Removed.

## Issue Type
- Bug

## Checklist

### DIFF
--- a/src/synthesizer/emission_models/stellar/models.py
+++ b/src/synthesizer/emission_models/stellar/models.py
@@ -394,12 +394,6 @@ class IntrinsicEmission(StellarEmissionModel):
             reprocessed (EmissionModel): The reprocessed model to use, if None
                 then one will be created.
         """
-        # Ensure what we've been asked for makes sense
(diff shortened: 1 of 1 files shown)


WillJRoper:
This was more about what intrinsic actually is. With `fesc=0` `Reporcessed=Intrinsic` while I have no issue removing it, it does mean that confusion will now be there.

christopherlovell:
I think the fact that this breaks when using the code flexibly (a core synthesizer principle) means it is more important than any potential confusion. I don't personally see any risk of confusion - yes intrinsic = reprocessed when fesc = 0, but that's fine, it is still true. Just because fesc = 0 doesn't mean a galaxy suddenly doesn't have an intrinsic spectra component.

WillJRoper (approved):

PR #847, opened 2025-02-20
Implementing the `fesc` Transformer

### LINKED ISSUE #832: Move fesc to a transform
Currently `fesc` is a hard coded argument on a spectra generation, but in an abstract sense applying an escape fraction is just a scalar transformation of an array. We should change all `fesc` instances to an emission model `Transform` derived class. This will also allow the implementation of an `fesc - 1` method, for obtaining the escaped and transmitted spectra, for example.

Will lead to breaking changes to all methods that currently use the `fesc` argument, but more flexibility going forward.

WillJRoper:
The treatment of escape fractions was a bit of a relic of times gone by. It used to be applied during extraction which posed some issues of its own. This PR moves the application of escape fractions to a `Transformation` operation with a corresponding transformer. 

- There is now an `EscapeFraction` and `EscapedFraction` pair of transformers (the latter being the inverse of the former). 
- `EscapeEmission` has been removed, instead tethering the escaped model to the transmitted model was originally intended. 
- Making a `TransmittedEmission` model now makes an extraction of the "full" transmitted emission, a "transmitted" model which is an `EscapeFraction` transformation and a "escaped" model which is a `EscapedFraction` transformation. 

Closes #832 

## Issue Type

- Enhancement

## Checklist


WillJRoper:
Ready now with AGN folded in too.

stephenmwilkins:
Approved, but I want to make sure it still does what's expected!

christopherlovell (approved):
As discussed offline, just need to change the name of the emission models to `EscapedFraction` and `ProcessedFraction`, then good to go

WillJRoper:
I've address the above comment. Just to be clear, should anyone look at this in the future the necessary change was `EscapeFraction` -> `ProcessedFraction`. This means:

`EscapedFraction` - is emission which escapes without reprocessing, i.e. `fesc * spectra`
`ProcessedFraction` - is emission which does not escape and is reprocessed, i.e. `(1 - fesc) * spectra`

PR #848, opened 2025-02-21
Introducing some simple caching of weights grids for integrated spectra

WillJRoper:
We have previously never held onto the calculated weights grids when extracting spectra for a number of reasons (memory footprints, complexity with escape fractions and masks etc). However, now we have escape fractions decoupled from extractions this is easier to navigate. 

This PR introduces caching of weight grids on a component, however, the weights are only cached and subsequently reused when not in the presence of a mask. This should reduce some overhead in duplicated calculations. 

This PR also includes a test of direct extraction from Python objects in the C extension, to remove the overhead introduced by the "prepare" methods. 

## Issue Type

- Enhancement

## Checklist


WillJRoper:
Changing tack... this is changing far too much at once!

PR #855, opened 2025-02-25
Use gnu99 standard on Linux

WillJRoper:
Compile with std=gnu99 on Linux. This deals with an error observed on cosma using gcc 14.1.0 where `posix_memalign` wasn't defined, simplest solution was to switch standard. 

Note that both c99 and gnu99 work on mac (darwin).

## Issue Type

- Bug

## Checklist


WillJRoper:
This PR is no out of sync because of the history changes.

PR #858, opened 2025-02-26
Laying the ground work for the new line format

WillJRoper:
This PR lays the groundwork for working with the new grid files. All this does so far is translate the new grid file format for lines into the old dictionaries, including full backward compatibility with old grids. 

Any further updates will remove all backward compatibility since we will be looking towards using all the new attributes and data.

## Issue Type

- Enhancement

## Checklist


WillJRoper:
Note that the new grid files as they stand have slightly bugged continua entries, because of this we still derive these from the spectra at read time. This isn't a big problem in the grand scheme but probably worth reading the data in the long run if we can.

stephenmwilkins (approved):

PR #867, opened 2025-03-06
Update dust curve defaults to calzetti20000

christopherlovell:
DWISOTT. It is very annoying to have to set a dust curve every time you load a default model. It is even more annoying having to do it twice for ISM and birth cloud attenuation (having different curves for each is quite a niche use case).

Also tidies some of the docstrings.

## Issue Type
- Enhancement

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced improved default settings for stellar emission models, including new dust curve options to enhance overall consistency and usability.
  - Added a comprehensive test suite for various emission models, validating initialization and functionality.

- **Bug Fixes**
  - Enhanced error messages for better clarity when parameters are not found, aiding in debugging and user understanding.

- **Documentation**
  - Expanded parameter descriptions to provide clearer guidance on default values, ensuring an intuitive user experience.
  - Updated pytest configuration for more detailed error tracebacks during test failures.

### DIFF
--- a/tests/test_premade_emodels.py
+++ b/tests/test_premade_emodels.py
@@ -48,7 +48,7 @@ class TestPacmanEmission:
         # Try to generate spectra without setting tau_v on the model or
         # stars
         random_part_stars.tau_v = None
-        with pytest.raises((RuntimeError, exceptions.MissingAttribute)):
+        with pytest.raises(exceptions.MissingAttribute):
             random_part_stars.get_spectra(model)
 
     def test_pacman_with_no_fesc(self, test_grid, random_part_stars):
@@ -428,7 +428,7 @@ class TestBimodalPacmanEmission:
 
         # Try to generate spectra without setting tau_v on the model or
         # stars
-        with pytest.raises((RuntimeError, exceptions.MissingAttribute)):
+        with pytest.raises(exceptions.MissingAttribute):
             random_part_stars.get_spectra(model)
 
     def test_bimodal_pacman_with_no_fesc(self, test_grid, random_part_stars):


src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -124,25 +125,33 @@ def __init__(
...
 
         Args:
-            grid (synthesizer.grid.Grid): The grid object.
-            tau_v (float): The V-band optical depth.
-            dust_curve (synthesizer.dust.DustCurve): The dust curve.
+            grid (synthesizer.grid.Grid):
+                The grid object.
+            tau_v (float):
+                The V-band optical depth. Defaults to 0.33.
+            dust_curve (synthesizer.dust.DustCurve):
```
WillJRoper, line 132: I know this is actually not something you have changed but the dust.DustCurve doesn't exist 😅 can you change this to `Transformer`.

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -112,6 +112,7 @@ class PacmanEmission(StellarEmissionModel):
     def __init__(
         self,
         grid,
+        tau_v=0.33,
```
WillJRoper, line 115: This shouldn't be set as a value. It will always override whats on the emitter. These parameters should only ever be set via kwargs that become the `fixed_parameters` dict on an `EmissionModel`

WillJRoper: i.e. you need to remove everywhere this has been passed. It has to be set on an emitter or override on a model. No parameter looked up with `get_param` can be set with a default argument on the model. This also simplifies models and means you have to adapt less places when changing things.

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -273,6 +282,7 @@ def _make_intrinsic_reprocessed(self, **kwargs):
     def _make_attenuated(self, **kwargs):
         return AttenuatedEmission(
             label="attenuated",
+            tau_v=self._tau_v,
```
WillJRoper, line 285: As above.

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -474,28 +484,37 @@ def __init__(
...
+            grid (synthesizer.grid.Grid):
+                The grid object.
+            tau_v_ism (float):
+                The V-band optical depth for the ISM.
+            tau_v_birth (float):
+                The V-band optical depth for the nebular.
+            dust_curve_ism (synthesizer.dust.DustCurve):
+                The assumed dust curve for the ISM. Defaults to `Calzetti2000`
+                with default parameters.
+            dust_curve_birth (synthesizer.dust.DustCurve):
```
WillJRoper, line 496: As above with type (my bad)

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -1252,8 +1271,8 @@ class CharlotFall2000(BimodalPacmanEmission):
         grid (synthesizer.grid.Grid): The grid object.
         tau_v_ism (float): The V-band optical depth for the ISM.
         tau_v_birth (float): The V-band optical depth for the nebular.
-        dust_curve_ism (synthesizer.dust.DustCurve): The dust curve for the
-            ISM.
+        dust_curve_ism (synthesizer.dust.DustCurve):
```
WillJRoper, line 1274: Here too but I'm going to stop highlighting this.

src/synthesizer/emission_models/stellar/pacman_model.py
```diff
@@ -1267,11 +1286,11 @@ class CharlotFall2000(BimodalPacmanEmission):
     def __init__(
         self,
         grid,
-        tau_v_ism,
-        tau_v_birth,
-        dust_curve_ism,
-        dust_curve_birth,
+        tau_v_ism=0.33,
```
WillJRoper, line 1289: Same here, these can't take non-string values.

WillJRoper (changes requested):
Mostly looks good and I definitely see the rationale about dust curves. See comments about optical depths though. These cannot be set to non-string defaults.

WillJRoper:
More general thing... we can add the default values for `tau_v` and the ism and birth cloud equivalents to a `StellarComponent`. This would achieve the default you were trying to set here. I could see the argument for doing that too. The pre made models then wouldn't need to state tau_v at all.

WillJRoper:
@coderabbitai review

src/synthesizer/emission_models/stellar/pacman_model.py
christopherlovell: Won't this be `synthesizer.emission_models.transformers.dust_attenuation.AttenuationLaw` ?

WillJRoper: Yea, but that is a child of `Transformer` and that's less of a meal. I'm fairly certain the API generator can work it out without the full path. `AttenuationLaw` is probably sufficient.

src/synthesizer/emission_models/stellar/pacman_model.py
christopherlovell: when you say 'the emitter' here, what are you referring to?

WillJRoper (approved):

PR #879, opened 2025-03-18
Fixing escape fractions in premade stellar models

### LINKED ISSUE #878: fesc doesn't properly modulate line emission
It appears that `fesc` doesn't properly modulate the nebular line/continuum emission, at least in the `IntrinsicEmission` model. In the example below I produce two spectra, one with `fesc=0` and one with `fesc=1`, which are the exact same. (note that I'm following the syntax in the [docs](https://synthesizer-project.github.io/synthesizer/emission_models/premade_models/stellar_models.html))

```python
from synthesizer.emission_models import IntrinsicEmission
from synthesizer.parametric import SFH, Stars, ZDist
from synthesizer.grid import Grid
from unyt import Msun, Myr

grid = Grid("test_grid", grid_dir="../synthesizer/tests/test_grid")
stellar_mass = 10**11 * Msun
sfh = SFH.Constant(max_age=100 * Myr)
zh = ZDist.DeltaConstant(metallicity=0.01)
stars = Stars(
    grid.log10age,
    grid.metallicity,
    sf_hist=sfh,
    metal_dist=zh,
    initial_mass=stellar_mass,
)
emission_model = IntrinsicEmission(grid, fesc=0)
spectra0 = stars.get_spectra(emission_model)

emission_model = IntrinsicEmission(grid, fesc=1)
spectra1 = stars.get_spectra(emission_model)

import matplotlib.pyplot as plt
fig, ax = plt.subplots()
ax.loglog(spectra0.lam, spectra0.lnu, label='fesc=0', alpha=0.5, color='r')
ax.loglog(spectra1.lam, spectra1.lnu, label='fesc=1', alpha=0.5, color='b')
ax.legend()
plt.show()
```
Image

I would expect that if `fesc=1`, all the stellar emission escapes _without_ photoionizing, and therefore there should be no nebular emission. But instead it seems to have no effect. 

The `fesc_ly_alpha` parameter seems to work just fine, though I'm not sure why we can't specify both (in `emission_models/stellar/models.py`:
```python
        # We can't define fesc and fesc_ly_alpha in the same model if
        # the user tried tell them they can't
        if "fesc" in kwargs:
            raise exceptions.InconsistentArguments(
                "Cannot define fesc and fesc_ly_alpha in the same model. "
                "Please use another Transformation on NebularLineEmission "
                "to apply your fesc."
            )
```

A more general question: what is the purpose of `fesc_ly_alpha`? If it is to allow the user to attenuate Lya emission, e.g. due to neutral gas, I would think it would be better served in an attenuation model, perhaps even in the IGM attenuation models. As it stands its a bit confusing, since fesc normally refers to ionizing photons escaping without contributing to nebular emission, while Lyman-alpha escape generally refers to Lya photons "escaping" without being attenuated by neutral gas.

WillJRoper:
The escape fraction was never actually being applied to the `NebularEmission` when set via a parent model such as `IntrinsicEmission`. This PR not only solves this issue but also adds some tests to ensure this works, both when a model is created with an attached escape fraction and when the `get_spectra` override argument is used.

Closes #878 

## Issue Type

- Bug

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Emission models now support a configurable escape fraction parameter, providing enhanced flexibility in generating nebular emissions.
  - Introduced a new pytest fixture for the `IntrinsicEmission` model to streamline testing.
- **Bug Fixes**
  - Refined parameter validation now offers consistent error handling for out-of-range emission fractions.
- **Tests**
  - Added validations confirm that varying the escape fraction produces distinct emission outputs, ensuring robust model behavior.
  - New test suite for the `transformers` module to verify the functionality of the escape fraction parameter across multiple models.

### DIFF
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -10,10 +10,12 @@ def test_fesc_model_level(
     random_part_stars,
 ):
     """
-    Test the fesc override in the emission model.
+    Test setting fesc on the emission model.
 
-    This test ensures that the fesc override on get_spectra methods is working
-    correctly.
+    This test ensures that the fesc set during instantiation of a model
+    is set correctly. We only do this for the IntrinsicEmission model here
+    since it encompasses all the models that relate to fesc with some parent
+    models derived from it.
     """
     # Get the spectra with fesc=0
     model = IntrinsicEmission(test_grid, fesc=0)
@@ -39,10 +41,11 @@ def test_fesc_override(
     pacman_emission_model,
 ):
     """
-    Test the fesc override in the emission model.
+    Test the get_spectra fesc override in the emission model.
 
     This test ensures that the fesc override on get_spectra methods is working
-    correctly.
+    correctly. Unlike above, we test all the models here to ensure the
+    override is correctly passed through different levels of complexity.
     """
     # For this test we need a singular tau_V value (we're only doing integrated
     # spectra)
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -1,0 +1,75 @@
+"""A test suite for the transformers module."""
+
+import numpy as np
+
+from synthesizer.emission_models import IntrinsicEmission
+
+
+def test_fesc_model_level(
+    test_grid,
+    random_part_stars,
+):
+    """
+    Test the fesc override in the emission model.
+
+    This test ensures that the fesc override on get_spectra methods is working
+    correctly.
+    """
+    # Get the spectra with fesc=0
+    model = IntrinsicEmission(test_grid, fesc=0)
+    spec_fesc0 = random_part_stars.get_spectra(model)
+    random_part_stars.clear_all_emissions()
+
+    # Get the spectra with fesc=1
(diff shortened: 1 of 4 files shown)


tests/test_transformers.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def test_fesc_override(
+    test_grid,
+    random_part_stars,
+    nebular_emission_model,
+    reprocessed_emission_model,
+    intrinsic_emission_model,
+    pacman_emission_model,
+):
```
hollisakins, line 40: Should the docstring for this test be different from the `test_fesc_model_level`?

hollisakins (changes requested):
Looks good, only thing is the docstring for `test_fesc_model_level` should be modified since it currently is the same as `test_fesc_override`, and I think its describing the latter.

tests/test_transformers.py
WillJRoper: D'oh, fixed!

hollisakins (approved):

PR #899, opened 2025-04-04
Adding a handler to ensure the C extensions aren't handed garbage

WillJRoper:
I and @marcoleonardi97 have separately had issues where simulation-specific readers are reading in float32 values but the C extension requires doubles (float64). Similarly, if you mask data then it is not contiguous (unless numpy makes a copy). Both of these can lead to garbage values but not an explicit failure. 

This PR introduces a helpful handler for the backend to do this conversion at the point of use. Importantly:
- We only want to do this conversion once. To make sure of this one of the helpers will get, convert and set if a conversion is needed, otherwise it just returns the value. 
- Only values that will actually be handed to the C are converted if needs be. 
- A user should never see any of this or need to care about this.

## Issue Type

- Bug
- Enhancement

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Stellar age and metallicity computations now produce results with enhanced precision by specifying data types.
  - Emission model parameters are processed with improved compatibility for more reliable numerical operations.
  - Additional utility functions enhance numerical handling and compatibility with C extensions, contributing to overall stability and accuracy.
  - New test suite introduced to validate the functionality of compatibility functions, ensuring robust performance.

- **Bug Fixes**
  - Improved handling of input data types for attributes in the `Particles` class, ensuring compatibility with C-style double arrays.

### DIFF
--- a/src/synthesizer/components/stellar.py
+++ b/src/synthesizer/components/stellar.py
@@ -88,7 +88,7 @@ class StarsComponent(Component):
             log10ages (array)
                 log10 stellar ages
         """
-        return np.log10(self._ages)
+        return np.log10(self._ages, dtype=np.float64)
 
     @property
     def log10metallicities(self):
@@ -99,7 +99,7 @@ class StarsComponent(Component):
             log10metallicities (array)
                 log10 stellar metallicities
         """
-        return np.log10(self.metallicities)
+        return np.log10(self.metallicities, dtype=np.float64)
 
     def __str__(self):
         """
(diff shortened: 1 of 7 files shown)


aswinpvijayan (approved):

aswinpvijayan:
I have no issues with this, but you would have already seen this, its going to increase our memory footprint a lot.

WillJRoper:
> I have no issues with this, but you would have already seen this, its going to increase our memory footprint a lot.

It isn't actually much of an increase really. We already assume double precision everywhere when things are allocated. It's only when data is fed in that the footprint could be increased.

There's a strong argument to allow the user to choose single or double precision globally though. These functions could be used to enable this with some minor modification

PR #905, opened 2025-04-17
Improvement to imaging performance

### LINKED ISSUE #889: Check image units when applying noise
The image units generated by galaxy.get_images_flux having been changed recently, some codes are at risk of facing unit issues if applying noise in the old units. To prevent this from happening if the units were to change again in the future and for the general case where the user makes a unit mistake, it would be beneficial to add a unit checking step in synthesizer when noise is applied to an image.

This feature could either raise an error to the user to tell them the noise units provided don't match the image units or internally convert the noise units to the image units (with the option of raising a warning to tell the user their noise units were wrong but got converted).

WillJRoper:
Imaging has come up as a bit of a bottle neck in recent production pipelines. While its a given that it should be expensive, it seemed there was still some performance left on the table. This PR applies some changes to the imaging machinery to this end (and also simplifies the backend for image generation):

- We allocate less often when looping over particles now.
- We call the extension once when generating multiple images. 
- We also collect together images for multiple different keys that can be generated at once on a model.
-  Image generation is now handled by a centralised set of generator function all defined in the same place. This simplifies maintenance where many different methods had ended up dotted around.

Closes #889.

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for generating multiple images simultaneously and improved handling of physical units in image and spectral cube creation.
  - Introduced new properties and methods for easier access to wavelength counts, IFU data, and photometry collection length.
  - Enabled assignment of images to collections using intuitive syntax.
  - Added a new exception for missing IFU data.

- **Improvements**
  - Streamlined and modularized image and spectral cube generation with dedicated helper functions for both particle-based and parametric emitters.
  - Enhanced placeholder instrument creation to better utilize available filter information.
  - Improved error messages for unit handling and input validation.
  - Optimized memory usage and parallel safety in image generation.

- **Refactor**
  - Refactored internal logic for image and spectral cube creation, delegating tasks to specialized helper functions and simplifying method signatures.
  - Updated C extension and Python interfaces for unified coordinate handling and multi-image support.
  - Simplified spectral cube constructor and improved argument validation.

- **Bug Fixes**
  - Ensured all wavelength arrays are unit-aware, improving compatibility and correctness.
  - Improved unit consistency checks when applying noise to images.


WillJRoper:
A quick addendum to this: This PR swaps the spectral data cube over to using the same machinery as the imaging. This seemed like a good idea, but after reflection, we actually want to explicitly sample the IFU wavelengths in the C extension for this to remove redundant big arrays during generation. This is fairly quick to change so I'll do it in another PR later.

src/synthesizer/base_galaxy.py
```diff
@@ -1103,8 +1103,28 @@ def get_images_luminosity(
             )
 
         # If we haven't got an instrument create one
+        # TODO: we need to eventually fully pivot to taking only an instrument
+        # this will be done when we introcued some premade instruments
```
aswinpvijayan, line 1107: typo

src/synthesizer/base_galaxy.py
```diff
@@ -1256,8 +1276,21 @@ def get_images_flux(
...
```
aswinpvijayan, line 1280: typo

src/synthesizer/components/component.py
```diff
@@ -499,8 +499,18 @@ def get_images_luminosity(
             )
 
         # If we haven't got an instrument create one
+        # TODO: we need to eventually fully pivot to taking only an instrument
+        # this will be done when we introcued some premade instruments
```
aswinpvijayan, line 503: typo

src/synthesizer/components/component.py
```diff
@@ -604,8 +614,18 @@ def get_images_flux(
...
```
aswinpvijayan, line 618: typo

src/synthesizer/emission_models/base_model.py
```diff
@@ -3011,45 +3016,114 @@ def _get_images(
...
+                    this_phot = (
+                        emitter.particle_photo_lnu[label]
+                        if this_model.per_particle
+                        else emitter.photo_lnu[label]
+                    )
+            except KeyError:
+                # Ok we are missing the photometry
+                raise exceptions.MissingSpectraType(
+                    f"Can't make an image for {label} without the photometry. "
+                    "Did you not save the spectra or produce the photometry?"
```
aswinpvijayan, line 3068: Is that too aggressive?

WillJRoper: That explicit error has been in the code base for a while. It's just spelling out exactly what the user didn't do. I've had various people say they're entertained by this tone 🤷‍♂️😂 so no, I don't think it's too aggressive and I will fight for my mildly passive aggressive error messages 😂

christopherlovell (approved):

PR #911, opened 2025-04-22
Add function to calculate tau at wavelength, not normalised to V-band

aswinpvijayan:
## Issue Type
- Enhancement

I have added a function for the dust attenuation curve to calculate the tau value at a wavelength, that is not normalised to the V-band. This has not been done for the ParametricLi08, since that form is fit to the normalised Alam/Av values and the paper does not give the observed Av for the different forms.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added the ability to directly compute optical depth at specific wavelengths for Calzetti2000, MWN18, and GrainsWD01 attenuation models.
  - Provided clear guidance on model-specific support for optical depth calculations.
  - Improved documentation for these new methods, including usage details and relevant citations.

### DIFF
--- a/src/synthesizer/emission_models/transformers/dust_attenuation.py
+++ b/src/synthesizer/emission_models/transformers/dust_attenuation.py
@@ -897,7 +897,6 @@ class ParametricLi08(AttenuationLaw):
             float/array-like, float
                 The optical depth.
         """
-
         raise exceptions.UnimplementedFunctionality(
             "ParametricLi08 form is fit to the normalised Alam/Av values"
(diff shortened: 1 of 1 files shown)


WillJRoper (changes requested):
LGTM. The only thing I would add is that the base `AttenuationLaw` should define a method that errors for this to ensure it's implemented on any future dust curves. For the `ParametricLi08` model, you can just implement the method but have it raise an error explaining why it doesn't make sense for that model.

WillJRoper (approved):

PR #922, opened 2025-05-05
Fully embrace copy-free ndviews when extracting from unyt objects

WillJRoper:
DWISOTT

On balance, a ~10^-8 precision difference is fine when the alternative is making massive copies anytime we want to do a unit conversion and get the value (which is very often!). 

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced unit conversion for arrays to reduce precision issues by performing in-place conversions.
  - Improved image addition compatibility by allowing approximate equality checks for resolution, field of view, and pixel counts.
- **Improvements**
  - Updated image resolution calculation to use fixed physical scale instead of relative fraction.
  - Refined pixel count computation by truncating instead of rounding up.
  - Adjusted image addition logic to accept nearly equal resolutions within floating-point tolerance.
- **Documentation**
  - Clarified expected minor precision differences and recommended equality checks for converted arrays.

### DIFF
--- a/docs/source/grids/grids.rst
+++ b/docs/source/grids/grids.rst
@@ -7,7 +7,7 @@ Introduction
 Most of the functionality of synthesizer is reliant on *grid files*. These are typically precomputed multi-dimensional arrays of spectra (and lines) from Stellar Population Synthesis (SPS) models for a range of ages and metallicities, and potentially other parameters (see below).
 Grids can also represent the emission from other sources, e.g. active galactic nuclei.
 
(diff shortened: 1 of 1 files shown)


src/synthesizer/imaging/base_imaging.py
```diff
@@ -126,7 +126,7 @@ def _compute_npix(self, compute_fov=True):
                 resolution and new npix. Defaults to True.
         """
         # Compute how many pixels fall in the FOV
-        self.npix = np.int32(np.ceil(self.fov / self.resolution))
+        self.npix = np.int32(self.fov / self.resolution)
```
aswinpvijayan, line 129: Why change from np.ceil?

WillJRoper: `np.ceil` actually introduces errors when the rounding is very close. It causes fewer issues when it's omitted. I will admit, though, I am not remotely happy with this and want to revisit it in the future.I just can't finding a way to parametrise it that doesn't lead to an edge case and this version fails the least.

aswinpvijayan (approved):

PR #927, opened 2025-05-13
Introducing some premade instruments and related machinery

### LINKED ISSUE #895: Provide some preconfigured instruments
In the spirit of removing as much friction as possible we should probably introduce some pre made instruments. I've found myself generating the same instrument sets over and over.

For instance, we could provide:
- NIRCam wide (all wide filters with a resolution)
- NIRCam ALL (all filters with a resolution)
- MIRI wide (ditto)
- JWST (Collection of everything)
- UVJ (rest frame instrument)
- Theoretical top hats (FUV NUV UV...)
- Euclid
- etc.

WillJRoper:
This PR introduces some premade instruments and therefore closes #895. These premade instruments define a lot of the an instrument's properties but also allow the user to set overrides and select subsets of filters from the predefined set. 

These premade instruments are implemented in much the same way as the premade emission models are (they are classes) since they can have various properties overridden. They can either be imported or loaded from cache files. A script for generating these cached files has been introduced in a separate directory in `src`, I'm not sold on this placement, but I'm not sure where else it should go. It doesn't really need to be run by the user, but it could be; it also requires extra dependencies. 

The documentation has been updated with a (small) new entry for the premade models.

The cached files can be downloaded from Box using the downloader. This has necessitated some serious updates to the downloader:
- A function now exists to probe the box and generate the database from the files in the box explicitly. This automates the whole downloader on the maintenance side! Like the instrument generation script, I'm not sure where the best place is for this. It requires a Box API and thus will never be run by a user. For now, it's importable from the `downloader` module.
- By default, the destination doesn't need to be stated now, with each file ending up in its expected default location without intervention.
- While doing this, I cleaned up the Box a bit.

These downloader updates could be split out into a separate PR, and maybe should b,e but the instruments needed them and it became clear that a bunch of finessing was necessary. I'm happy to spin it out if you want.

This isn't the end of premade models but this seemed a good well rounded start.  I would like to:
- Add a `TheoreticalTopHats` instrument defining all the rest frame top hats we used to use in an easy way. 
- Introduce baked-in PSFs to both the HST instruments and Euclid, but both are a bit more involved than using webbPSF.
- Introduce spectroscopic instruments (some ground work has been laid for this with a wavelength array generation utility added for converting from resolving power).

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive set of premade astronomical instruments (e.g., JWST, HST, Euclid) that can be easily imported and used, including utilities for listing and loading them.
  - Added command-line options and supporting logic to download instrument data files alongside existing datasets.
  - Provided a new documentation notebook and index entries explaining how to use premade instruments and manage instrument data.
  - Added a utility for generating wavelength arrays based on resolving power.

- **Bug Fixes**
  - Improved error handling when loading SVO filter data, ensuring clearer exceptions if filters are missing.

- **Documentation**
  - Expanded documentation with new guides and examples for premade instruments, including instructions for downloading and using instrument files.

- **Chores**
  - Added scripts and workflows for generating and managing instrument cache files and data download mappings.
  - Included README files to clarify instrument cache usage and data management.


src/synthesizer/downloader/downloader.py
```diff
@@ -170,6 +231,20 @@ def download_sc_sam_test_data(destination):
...
+    Args:
+        destination (str):
+            The path to the destination directory.
+        instruments (list):
+            The list of instruments to download.
+    """
+    # Download each instrument
+    for instrument in instruments:
+        _download(INSTRUMENT_TRANSLATION[instrument], destination)
+
```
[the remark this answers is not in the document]
WillJRoper: Not possible

.github/workflows/python-app.yml
```diff
@@ -50,8 +50,9 @@ jobs:
       - name: Download test data
         run: |
           # Download test grid data
-          synthesizer-download --test-grids -d tests/test_grid/ --dust-grid
-          synthesizer-download -d tests/data --all-sim-data
+          synthesizer-download --test-grids --dust-grid -d tests/test_grid/
+          synthesizer-download --all-sim-data -d tests/data/
+          synthesizer-download --instruments EuclidNISP
```
christopherlovell, line 55: Is the Euclid Instrument (/ PSF) lightweight enough for workflow / testing?

src/insturment_generation/get_instruments.py
```diff
@@ -0,0 +1,273 @@
+"""This script will generate all instrument cache files.
```
christopherlovell, line 1: typo in file / directory name

src/synthesizer/instruments/utils.py
```diff
@@ -0,0 +1,136 @@
+"""A submodule for instrument related utility functions.
+
+Example usage:
+
+    # Generate a wavelength array with a constant resolving power of 10000
+    lams = generate_wavelength_array(400, 700, 10000)
```
christopherlovell, line 6: this method does not exist in this file

christopherlovell (changes requested):

src/synthesizer/instruments/utils.py
WillJRoper: Fixed (I changed the function name but not the doc string)

src/insturment_generation/get_instruments.py
WillJRoper: Doh 😂 Fixed

christopherlovell (approved):

Issue #963, opened 2025-06-25
Line luminosities for "total" are zero and shouldn't be


stephenmwilkins:
At the moment the line luminosities for the Grid computed convenience `total` component are zero, while they should in fact be equal to `nebular`. I'm somewhat confused why this didn't emerge earlier, but maybe it simply reflects that total isn't routinely used?

This would be a simple fix, but, in fact, I wonder if we should deprecate `total` entirely since it implicitly assumes `fesc=0` and I don't think it's used much? Instead people should generate combined spectra using an emission model where it is clear what it is doing. Alternatively `total` could be renamed something else.

WillJRoper:
I don't think `total` is ever used 😂 and you are right, it serves no purpose in context now. Lets stamp it out!

christopherlovell:
Raised in #977, moving discussion there.

PR #970, opened 2025-07-21
Fix: Set total line luminosities equal to nebular (#963)

christopherlovell:
## Summary

Fixes issue #963 where line luminosities for the "total" component were incorrectly set to zero when they should equal nebular line luminosities.

### Background
The "total" spectra correctly combines `transmitted + nebular` spectra, providing a convenient way to get the total emission (ignoring dust reprocessing). However, the corresponding line luminosities were incorrectly being set to zero for the "total" component.

### Root Cause
In `Grid._get_lines_grid()`, all non-native spectra types (anything other than nebular, transmitted, etc.) were having their line luminosities set to zero by default. This was incorrect for "total" since:
- Total spectra = transmitted + nebular spectra ✅ (correct)
- Total line luminosities = 0 ❌ (incorrect)
- Should be: Total line luminosities = nebular line luminosities ✅ (since transmitted lines = 0)

### Changes Made
- Modified `_get_lines_grid()` method in `src/synthesizer/grid.py` (lines 640-641)
- Added special case handling for "total" spectra type
- Set `line_lums["total"] = line_lums["nebular"]` instead of zeros
- Maintains consistency with how total spectra are computed

### Testing
- ✅ Existing test `tests/test_grid.py::TestSPSGridLines::test_lines_different` passes
- ✅ All grid tests continue to pass
- ✅ Verified total line luminosities now equal nebular values (9.93e+31 erg/s)
- ✅ No breaking changes to existing functionality

### Code Diff
```python
# Before
self.line_lums[spectra] = unyt_array(np.zeros(...), lum_units)

# After  
if spectra == "total":
    self.line_lums[spectra] = self.line_lums["nebular"]
else:
    self.line_lums[spectra] = unyt_array(np.zeros(...), lum_units)
```

This fix ensures that total line luminosities are consistent with the physics: since transmitted line emission is zero by definition, total line emission should equal nebular line emission.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted how line luminosities are initialized for the "total" spectra, ensuring they now match the "nebular" spectra instead of being set to zero. This improves the accuracy of displayed line luminosities for the "total" spectra type.

### DIFF
--- a/docs/source/emission_grids/grids_example.ipynb
+++ b/docs/source/emission_grids/grids_example.ipynb
@@ -104,7 +104,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "grid = Grid(\"test_grid.hdf5\", grid_dir=\"../../../tests/test_grid\")"
+    "grid = Grid(\n",
+    "    \"test_grid.hdf5\", grid_dir=\"../../../tests/test_grid\", ignore_lines=True\n",
+    ")"
    ]
   },
   {
@@ -175,7 +177,11 @@
     "new_lams = np.logspace(2, 5, 1000) * angstrom\n",
     "\n",
     "# Create a new grid\n",
-    "grid = Grid(\"test_grid\", new_lam=new_lams)\n",
+    "grid = Grid(\n",
+    "    \"test_grid\",\n",
+    "    # ignore_lines=True,\n",
+    "    new_lam=new_lams,\n",
+    ")\n",
     "print(grid.shape)"
    ]
   },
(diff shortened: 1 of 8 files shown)


WillJRoper:
The conclusion on #963 was actually just that `"total"` served no purpose and should be removed rather than fixed. It's also confusing in that "total" elsewhere in the code includes not only dust extinction but dust emission too so having total here be some "total intrinsic" named "total" only adds confusion. 

What do you think? Shall we kill it off or rename it? My vote is the former since it serves no utility and hasn't for an age (if ever).

christopherlovell:
> The conclusion on #963 was actually just that `"total"` served no purpose and should be removed rather than fixed. It's also confusing in that "total" elsewhere in the code includes not only dust extinction but dust emission too so having total here be some "total intrinsic" named "total" only adds confusion.
> 
> What do you think? Shall we kill it off or rename it? My vote is the former since it serves no utility and hasn't for an age (if ever).

Agreed. Since this is a more fundamental change, I'll close this and raise a new issue.

PR #971, opened 2025-07-21
Update mask attribute assignment

christopherlovell:
Ensures masks are properly initialised in an emission model, and used to update naming scheme.

## Issue Type
- Bug

## Checklist

### DIFF
--- a/src/synthesizer/emission_models/base_model.py
+++ b/src/synthesizer/emission_models/base_model.py
@@ -310,11 +310,11 @@ class EmissionModel(Extraction, Generation, Transformation, Combination):
         # Are we making per particle emission?
         self._per_particle = per_particle
 
-        # Define the container which will hold mask information
-        self.masks = []
-
-        # If we have been given a mask, add it
-        self._init_masks(mask_attr, mask_thresh, mask_op)
+        # Save the mask info and initialise
+        self.mask_attr = mask_attr
+        self.mask_thresh = mask_thresh
+        self.mask_op = mask_op
+        self._init_masks()
 
         # Initialise the lam mask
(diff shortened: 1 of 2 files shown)


src/synthesizer/emission_models/base_model.py
```diff
@@ -310,11 +310,11 @@ def __init__(
         # Are we making per particle emission?
         self._per_particle = per_particle
 
-        # Define the container which will hold mask information
-        self.masks = []
-
-        # If we have been given a mask, add it
-        self._init_masks(mask_attr, mask_thresh, mask_op)
+        # Save the mask info and initialise
+        self.mask_attr = mask_attr
```
WillJRoper, line 314: What's the rationale for having these attached to a model? A model can contain multiple masks, each of whicu can be added via the setter, so attaching a particular attribute and threshold is going to be a bit confusing.

src/synthesizer/emission_models/base_model.py
```diff
@@ -718,9 +719,16 @@ def _unpack_model_recursively(self, model):
             # The model is already in the tree so nothing to do
             pass
         else:
-            raise exceptions.InconsistentArguments(
-                f"Label {model.label} is already in use."
-            )
+            # Ensure model has a unique name
+            if self.mask_attr is not None:
+                model.label += f' ({mast_attr} {mask_op} {mask_thresh})'
```
WillJRoper, line 724: This is a nice solution though. Can't we keep this behaviour and nix the new mask attributes? 

Previously we've solved this issue by just adding modifiers in the pre-made models (young/old etc) that avoid the name clash error that can come up.

PR #991, opened 2025-08-13
Supporting surviving stellar mass calculations for parametric & particle galaxies

tHarvey303:
I used the existing functionality in the grid_generation repo to add the 'stellar_fraction' and 'remnant_fraction' arrays to a BPASS grid. This should be possible to add to all existing BPASS grids quite easily, and also for the other SPS grids. 

I've got a simple base implementation of enabling `Grid` to store the stellar fraction, and corresponding methods on particle `Stars` and parametric `Stars` to calculate the surviving stellar mass when given the grid. I've also added a corresponding convenience function to `BaseGalaxy` to do the same calculation at a galaxy level. 

In the case where there is no stellar fraction stored on the grid, it is set to unity for all ages and metallicities, so the methods will return simply the total initial mass. This could be implemented differently however.

The implementation works currently for both parametric and particle galaxies. It currently uses the `get_sfzh()` method on the particle galaxy, and then simple sums the multiplication of the two arrays. I'm however unsure if the `get_sfzh()` method will always provide a correctly normalzied `sfzh` array if `initial_masses` and `current_masses` differ, or some of the particles are replaced with parametric stars, so this may need changing (and is why I've implemented the methods seperately, rather than on the parent `StarsComponent`). 

### Example Usage

```  python
grid = Grid('bpass-2.2.1-bin_chabrier03-0.1,300.0.hdf5',
            grid_dir='/home/tharvey/work/synthesizer_grids/generated/')

sfh = SFH.LogNormal(peak_age=1*Myr, tau=0.3, max_age=100*Myr)

galaxy = Galaxy(
    redshift=1,
    stars = Stars(
        grid.log10age,
        grid.metallicity,
        sf_hist = sfh,
        initial_mass =1e9 * Msun,
        metal_dist = ZDist.DeltaConstant(log10metallicity=-1.5),
        tau_v= 0.2,
    ),
)

galaxy.get_surviving_mass(grid)

# or

galaxy.stars.calculate_surviving_mass(grid)

-------------------------------
unyt_quantity(8.29210115e+08, 'Msun')
```

I'm sure people will have opinions about this so I'm creating this PR as an initial idea for disscussion. I haven't implemented tests or updated docs for this as the implementation might change (and the test grids don't support this).


src/synthesizer/base_galaxy.py
```diff
@@ -546,6 +546,10 @@ def get_photo_fnu(self, filters, verbose=True, nthreads=1):
                 nthreads=nthreads,
             )
 
+    def get_surviving_mass(self, grid):
```
WillJRoper, line 549: Can we call it `get_remenant_mass`? Or at the very least provide that as an alias.

src/synthesizer/grid.py
```diff
@@ -465,6 +467,28 @@ def _get_ionising_luminosity(self):
...
+        """Get the stellar fraction from the HDF5 file.
+
+        This is a two-dimensional array with the first axis being age and
+        the second axis being metallicity.
+        """
+        with h5py.File(self.grid_filename, "r") as hf:
+            if key in hf.keys():
+                self.stellar_fraction = hf[key][:]
+            else:
+                self.stellar_fraction = np.ones(self.shape[:2])
```
WillJRoper, line 480: This should error I think. It does mean you can't jsut call the loader in the init though. Maybe instead implement `stellar_fraction` as a property method that caches the stellar fractions in a private attribute and returns it if it exists, otherwise it throws an error? 

Fundamentally, this data does not exist and I could see some blindly using this with a grid without it and not realising.

src/synthesizer/base_galaxy.py
tHarvey303: Can do, but might be unclear at least for BPASS - remnant_fraction (i.e. the fraction of the total mass in stellar remnants - NS, WD, BHs) is also stored on the grid so it might not be clear exactly what 'remnants' is referring to. We could also have a separate method to calculate the mass in remnants (I don't know if this is available for all grids though).

WillJRoper:
This should be good to go in really, just needs the formatter run on it. Feel free to undraft when you are happy for us to have another look.

tHarvey303:
Should be good to go now. Rerunning existing grids shouldn't be too hard, not sure if there is an existing method to add them to already post-processed grids within the grid generation repo.

src/synthesizer/grid.py
```diff
@@ -174,6 +174,9 @@ def __init__(
         self._model_metadata = {}
         self._get_grid_metadata()
 
+        # Set the internal flags
```
WillJRoper, line 177: Nitpicky, but this isn't a flag, its an unintialised caching attribute. Be clear about that in the comment so the next maintainer knows at a glance.

PR #1002, opened 2025-09-11
Modify disc transmission model

stephenmwilkins:
In the incarnation of this model the transmitted disc spectra was the combination of the escaping light and then the sum of the transmission through the NLR and BLR weighted by the covering fraction. However, this does not make much sense since an observer is only seeing one line of sight. 

Instead this PR replaces this combination with a choice between being transmitted through the BLR, NLR, or simply escaping. This is decided by the `disc_transmission` parameter which can be either `nlr, blr, none, or random`. 

At the moment this implementation will only work for a single BH so will need to be vectorised somehow. I invite ideas.

## Issue Type
- [x] Enhancement

## Checklist

### DIFF
--- a/src/synthesizer/photoionisation/cloudy17.py
+++ b/src/synthesizer/photoionisation/cloudy17.py
@@ -190,7 +190,7 @@ def create_cloudy_input(
             cinput.append(
                 (
                     f"element abundance {abundances.element_name[ele]} "
-                    f"{abundances.total[ele]} no grains\n"
+                    f"{abundances.gas[ele]} no grains\n"
                 )
             )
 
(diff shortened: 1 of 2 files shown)


WillJRoper (changes requested):
Beyond the comment, this needs to have the combined version added.

stephenmwilkins:
I think more fundamentally the preferred "random" option won't work for an array of blackholes right? In that case you want each black hole to be independent - i.e. some would go through the NLR, some the BLR, and some escape completely. It's not obvious to me the best way of implementing this.

You could use random choice to populate an array of options e.g. np.array(['blr', 'blr', 'nlr', 'blr', 'escape']) and then use a loop but that's horrible.

The other option would be to set a weight for the blr, nlr, escape, and add them together. This then naturally works for the combined/averaged case.

where transmission = blr, you'd then have weight_blr = 1, weight_nlr=0, weight_escaped=0. 

What do you think?

PR #1006, opened 2025-10-05
Improve unified agn

stephenmwilkins:
At the moment, `UnifiedAGN` can't take independent parameters for BLR and NLR. This PR addresses this. 

## Issue Type
- Bug

## Checklist

### DIFF
--- a/docs/source/emissions/spectra/blackholes.ipynb
+++ b/docs/source/emissions/spectra/blackholes.ipynb
@@ -20,7 +20,7 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
-    "from unyt import K, Mpc, Msun, deg, yr\n",
+    "from unyt import K, Mpc, Msun, cm, deg, yr\n",
     "\n",
     "from synthesizer import Grid\n",
     "from synthesizer.emission_models import (\n",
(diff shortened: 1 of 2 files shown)


WillJRoper (changes requested):
All fine, just need to address the escaped confusion.

stephenmwilkins:
For the time being though `disc_escaped` actually makes sense as a label though. When looking through the torus there is no `disc_escaped` emission that we see hence why it gets set to zero.

The confusion arises because for the emission not transmitted through the torus we are combining, which as I mentioned I don't think makes sense.

stephenmwilkins:
So, essentially the labelling is fine at the moment, but in a new branch I'll investigate that more substantial change.

aswinpvijayan:
> Let me think about this a bit more, as it overlaps with another change I’ve been considering.
> 
> I think we should handle a few distinct scenarios:
> 
> * High inclination: If the inclination is too high, the line of sight passes through the torus. In this case, we can simply set disc_transmitted = 0.
> * Low inclination: If the inclination is low enough, there are three possible sight-lines:
> 
> 1. Through the NLR
> 2. Through the BLR
> 3. Through nothing (in which case we see the pure incident emission)
> 
> Currently, our approach combines the escaped emission with the BLR and NLR transmitted components, weighted by their covering fractions. However, as mentioned earlier, I don’t think this is physically consistent — we only observe a single sightline, not an average over all possible orientations.
> 
> That said, I do think we should also report an averaged disc emission quantity, `disc_isotropic`, which accounts for transmission through the different components according to their relative covering fractions.
> 
> @aswinpvijayan any thoughts?

You are talking about the isotropic model for emission here, right?

Also since we do not resolve any of these components, or have much information from the sims to constrain this, wouldn't having the option to use the average make sense, rather than imposing some arbitrary orientation? Also isnt the disc emission need to be an average across all orientation anyway?

WillJRoper (approved):

PR #1015, opened 2025-10-14
Removing SFZH attr on parametric `Galaxy`

WillJRoper:
Parametric galaxies were carrying a vestigial attribute holding a direct pointer to the SFZH. This was only there for `get_ionising_photon_luminosity`, which should have been on the component anyway. I've fixed this and also removed the long-deprecated `get_logQ` method.

## Issue Type

- Bug

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Clearer validation error when a required grid property is missing.
  * Ionising photon luminosity can now be obtained directly from star models.

* **Refactor**
  * Deprecated ionising-photon accessors on galaxy models removed to streamline the API.

### DIFF
--- a/src/synthesizer/parametric/stars.py
+++ b/src/synthesizer/parametric/stars.py
@@ -952,22 +952,30 @@ class Stars(StarsComponent):
 
         return surviving_mass * Msun
 
-    def get_ionising_photon_luminosity(self, grid: Grid) -> float:
+    def get_ionising_photon_luminosity(
+        self,
+        grid: Grid,
+        ion: str = "HI",
+    ) -> float:
         """Calculate the ionising photon luminosity from the grid.
 
         Args:
             grid (object, Grid):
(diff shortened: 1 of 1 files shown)


src/synthesizer/parametric/stars.py
```diff
@@ -951,3 +951,23 @@ def calculate_surviving_mass(self, grid: Grid):
         surviving_mass = np.sum(self.sfzh * grid.stellar_fraction)
 
         return surviving_mass * Msun
+
+    def get_ionising_photon_luminosity(self, grid: Grid) -> float:
```
christopherlovell: Can this be generalised to different ionising energies?

WillJRoper: The old version this is ported from didn't but there are other ions on the `Grid` (but grid dependent). I've added that as a kwarg

PR #1018, opened 2025-10-15
Caching cosmology calculations to remove overhead

WillJRoper:
It turns out that the observed spectra calculation and a few other repeated calculations can become disproportionately expensive because of the behind-the-scenes integration done during astropy cosmology calculations. We can't provide cheap look up tables for this because cosmologies can vary, so instead I've implemented a "simple" `lru_cache` wrapper implementation around these calculations. This will cache results and immediately return previously computed results. 

The reason I say "simple" is that `astropy` cosmology objects are not hashable, so I've had to chuck in some hash key magic to get around that fact, but this is safe and cheap.

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added cached cosmology distance utilities for faster, consistent distance lookups.

* **Bug Fixes**
  * Fixed unit handling for distance quantities and preserved correct zero-redshift behavior.

* **Refactor**
  * Centralized distance computations across distance, conversion, emission, and component logic.

* **Tests**
  * Added comprehensive tests for distance values, units, relations, multi-cosmology scenarios, redshift edge cases, and caching.


WillJRoper:
Ah damn... not as much of a saving as I had hoped. Clearly in your case the time isn't spent in the cosmology calculation :|

christopherlovell (changes requested):

Issue #1010, opened 2025-10-08
Global galaxy population parent class


christopherlovell:
For population studies, it is often the case that you wish to define a set of global parameters or hyperparameters that control the downstream properties of galaxies. For example, in the continuity model you may wish to set some prior on the student's t-distribution that describes how bursty or smooth the overall SFH is. For population inference you may then wish to constrain these parameters from the properties of the individual objects as well as the population distribution.

To enable these kinds of tasks, @tHarvey303 and I discussed creating a `GalaxyPopulation` parent class, that defines a few of these parameters, and allows users to generatively create galaxies with properties sampled from these distributions. 

Some open questions include:
- where should these generative methods be defined? (since they will be parametric SFH-specific, they should probably be on each parametric SFH object, but it would be nice to have a unified parent method that standardises this). 
- How should the generated galaxies be stored? _should_ they be stored anywhere (or just created at runtime)?
- How should such a class interface with e.g. a lightcone? should we have a single object instead of multiple?

PR #1022, opened 2025-10-20
Add new non-parametric SFHs and average star-formation rate method

tHarvey303:
Two major additions here:
1. 4 new 'non-parametric' SFHs - 'continuity', 'continuity-flex', 'dirichlet' and 'continuityPSB'. Primarily based on these SFHs as implemented in [Prospector](https://prospect.readthedocs.io/en/stable/sfhs.html). They are all time-binned SFHs. 
   - 'Continuity' -The SFR of each bin is derived from sampling a vector of parameters describing the ratio of SFRs in adjacent temporal bins. By default, a Student-t prior distribution is placed on the log of these ratios. Lower student-t concentration parameters will produce burstier SFHs. See Leja et al. 2017, 2019
   - 'ContinuityFlex' - The edges of the temporal bins are adjusted such that for a given set of SFRs an equal amount of mass forms in each bin. See Leja et al. 2019.
   - 'Dirichlet'- Dirichlet-prior binned SFH where the time bins are fitted for too ensure equal quantiles in each lookback bin.
   - 'ContinuityPSB' - This parameterization provides a number of fixed width bins at both small and large lookback times, combined with a number of flexible width bins between these fixed bins. Designed for post starburst galaxies, see Suess et al. 2021

As these star-formation histories are useful primarily because of the prior distributions relating the SFRs/widths of different bins of star formation, I have added a new classmethod called ```init_from_prior``` to them to draw a random SFH from the prior given the required parameters (e.g. bins of lookback time, redshift, concentration parameter, etc). I did this a while ago, and we may wish to move or have similar functionality on a ```GalaxyPopulation``` level (see #1010), but it is also useful to have on the level of an individual star-formation history. 

I've also added a ```CombinedSFH```, where you can use a weighted combination of different parametric SFHs. You can't add them directly as they are unnormalized, but you can use this to build more complex parametric SFHs if required.

2. A new ```calculate_average_sfr``` method. This is a Stars level method to calculate actual SFRs. The existing SFR methods for parametric SFHs aren't really suitable as they aren't normalized by the stellar mass (which is stored on the Stars level) and they don't deal with different time intervals well. I've added a new method which  can be used to determine the average SFR across any time range and which deals with partially overlapping SFH bins.

**I'm mainly opening this so we can discuss the best way to implement SFHs like these into Synthesizer, and there may be a better way to do it than this. I also plan to implement the SFH Power Spectrum model of Wan et al. 2024, which will be particularly useful for sampling a population of SFHs. Comments/thoughts welcome!** 

## Summary by CodeRabbit

* **New Features**
  * Vectorised SFR evaluation for batch age inputs
  * Initialize SFH models from priors for sampling workflows
  * Combine multiple SFH models into weighted composite SFHs
  * New average SFR calculator with flexible time-range input and unit-aware output
  * Accessors to retrieve SFR and mass fractions from non-parametric models
  * Plotting now labels age and normalized SFR axes

* **Bug Fixes / Validation**
  * Validation to detect inconsistent SFH grid values

* **Documentation**
  * Added parametric SFH docs and an example notebook demonstrating models, priors, plotting and composition

### DIFF
--- a/src/synthesizer/parametric/sf_hist.py
+++ b/src/synthesizer/parametric/sf_hist.py
@@ -1406,7 +1406,7 @@ class ContinuityFlex(Common):
     def _logsfr_ratios_to_agebins(self):
         """Transform SFR ratios to age bins assuming constant mass per bin."""
         logsfr_ratios = np.clip(self.logsfr_ratios, -10, 10)
-        _, y1 = 10 ** self.fixed_young_bin[0], 10 ** self.fixed_young_bin[1]
+        y0, y1 = 10 ** self.fixed_young_bin[0], 10 ** self.fixed_young_bin[1]
         o0, o1 = 10 ** self.fixed_old_bin[0], 10 ** self.fixed_old_bin[1]
         if y1 >= o0:
             raise ValueError("Young bin must end before start of old bin.")
@@ -1421,9 +1421,9 @@ class ContinuityFlex(Common):
         )
         dt1 = tflex / denom
 
-        # Build absolute age edges: [0, y1] (young), then flexible bins,
+        # Build absolute age edges: [y0, y1] (young), then flexible bins,
         # then [o0, o1] (old)
-        edges = [0.0, y1]
+        edges = [y0, y1]
         last = y1 + dt1
         edges.append(last)
         for i in range(n_ratio):
--- a/src/synthesizer/parametric/stars.py
+++ b/src/synthesizer/parametric/stars.py
@@ -939,7 +939,7 @@ class Stars(StarsComponent):
 
         return weighted_mean(attr, weight)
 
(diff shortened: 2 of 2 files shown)


tHarvey303:
Might as well see what the rabbit thinks anyway: @coderabbitai review

WillJRoper (changes requested):
Beyond the comment above I'm pretty happy to see this merged in this guise. 

It needs some clean up in the documentation (there's some rogue numpy style doc strings and some more comments needed in places) Other than that the only issues I can really level is that:
- We need to add the new SFHs to the docs (though I think this actually motivates splitting up the SFH section a bit more!)
- We really need to add some test coverage for this module but since this is missing currently we can do this in a subsequent PR. The docs will test the functionality you're adding well enough.

christopherlovell (approved):

PR #1034, opened 2025-10-31
Minor fix to warning formatter to stop crash when running in slurm/detached terminal

tHarvey303:
I've had this warning formatter occasionally cause a crash when running a script in slurm or a detached terminal. Just added a simple check to avoid negative widths. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when using small terminal windows by preventing edge case errors during text formatting.


WillJRoper (approved):
I found this and fixed it on a branch I never raised 😅

PR #1039, opened 2025-11-05
wrong warning in filter.py/apply_filter when passing galaxy with zero stars

aswinpvijayan:
Closes issue 1038

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced filter operation warnings. The system now provides more targeted feedback by only alerting users when filtering produces unexpected empty results. Warnings for empty input arrays are suppressed, reducing noise and allowing users to focus on genuine filtering concerns.

### DIFF
--- a/src/synthesizer/instruments/filters.py
+++ b/src/synthesizer/instruments/filters.py
@@ -1783,7 +1783,8 @@ class Filter:
 
         # Warn and exit if there are no array elements in this band
         if arr_in_band.size == 0:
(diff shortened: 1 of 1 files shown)


WillJRoper:
For posterity: the issue here is that the check for a filter being outside the input array just checked that the masked input (derived from non-zero transmission) was not zero in size. However, if the input array was zero-sized (0 stars) then this also triggers as a false positive.

WillJRoper (approved):

PR #1066, opened 2025-11-27
Handling smoothed imaging edge cases and expanding tests

WillJRoper:
@tHarvey303 found that there was a sweet spot of combined pixel resolution and smoothing kernel size that led to particles not contributing to the image.

It turns out this was due to the naive treatment of kernels in the imaging where pixels were evaluated at the centre which for small kernels may have left all pixel centres outside the particle's kernel despite the pixel overlapping with the kernel. This approach was cheap but too inaccurate to keep. 

We instead now treat each particle either as wholly within the kernel and do a simple average over the pixel area, or adaptively evaluate the kernel across the pixel to find the fraction inside the kernel and thus find the accurate contribution in all cases. This is an increase in the cost of the calculation, but it is a necessary one to ensure a correct result. The test suite has been expanded to verify this.

This also addresses a tiny edge case where the vmin and vmax derivation would fail for singularly populated pixels or when the majority of the pixels are 0. 

## Issue Type

- Bug

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flux conservation in smoothing via normalized kernel contributions; better visualization of sparse images by excluding zeros from percentile normalization.

* **Removed Features**
  * The compiled spectral-cube extension and its Python-facing APIs have been removed.

* **Behavior Changes**
  * Centralized centering validation for image generation now consistently warns or errors on mis-centered coordinates.

* **Tests**
  * Extensive new tests for flux conservation, pixel-overlap, threading consistency, and multi-filter imaging.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/docs/source/emissions/emission_objects/sed_example.ipynb
+++ b/docs/source/emissions/emission_objects/sed_example.ipynb
@@ -981,6 +981,82 @@
     "# Get attenuation at 5500 angstrom\n",
     "att_at_5500 = get_attenuation_at_5500(int_sed, att_sed)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9d38967c",
+   "metadata": {},
+   "source": [
+    "### Doppler broadening\n",
+    "\n",
+    "`Sed` objects also provide tools for producing Doppler-broadened versions of an SED. The method `doppler_broaden` returns a new SED broadened using a specified velocity dispersion. This can either operate in-place or return a new `Sed` object.\n",
+    "\n"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -303,19 +303,6 @@ def create_extension(
         links=link_args,
         include_dirs=include_dirs,
     ),
-    create_extension(
```
WillJRoper: This extension is long redundant code and should have already been removed.

PR #1078, opened 2025-12-15
Add GALEX premade instruments

sophie-newman:
Since SVO returns effective area rather than transmission curves for GALEX, I have defined the NUV and FUV transmission curves and created GALEXFUV and GALEXNUV Instruments within `instruments/premade.py` with their corresponding values of resolution. 

GALEXFUV and GALEXNUV are then combined into a single `InstrumentCollection`, which allows GALEX to be imported in the following way:

```
>>> from synthesizer.instruments import GALEX
>>> galex = GALEX()
>>> for inst in galex:
...     print(inst.label, inst.resolution)
... 
GALEXFUV 6 arcsec
GALEXNUV 8 arcsec
```

The only thing I don't love is that when you do the following

```
from synthesizer.instruments import print_premade_instruments
print_premade_instruments()
```

GALEXFUV and GALEXNUV are shown, and not GALEX. But this is because GALEX is not a `PremadeInstrument` and is instead an `InstrumentCollection` made of two `PremadeInstrument`s.

## Issue Type
- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a factory base for premade instrument collections and new GALEX premade instruments (FUV — 6", NUV — 8"); added GALEX, Euclid, HST, and JWST collection factories.

* **Bug Fixes**
  * Emission-line ratio now returns scalars for scalar results and arrays otherwise.

* **Chores**
  * Instrument listing now recognizes collection factories; public API expanded to expose new collections. Numerical integration calls updated; CI dependency source changed.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/synthesizer/instruments/filters.py
+++ b/src/synthesizer/instruments/filters.py
@@ -1632,7 +1632,10 @@ class Filter:
                     "can sometimes occur where the effective "
                     "area is returned instead. Please check "
                     "that the filter you are querying returns "
-                    "the transmission / response."
+                    "the transmission / response. (Note that for the case "
+                    "of GALEX we provide an importable instrument that "
+                    "handles this correctly: "
+                    "`from synthesizer.instruments import GALEX`.)"
                 )
             )
 


christopherlovell:
Given that these are essentially the same 'instrument' I would prefer to have a GALEX instrument that contains both filters. This should be possible (see EuclidNISP)

sophie-newman:
> Given that these are essentially the same 'instrument' I would prefer to have a GALEX instrument that contains both filters. This should be possible (see EuclidNISP)

Agreed but the FUV and NUV have different resolutions - any thoughts on how to go about this?

WillJRoper:
Note that as well as adding the `GALEX` collection, this introduces a second base class for collections of premade instruments.

christopherlovell (approved):

PR #1129, opened 2026-04-20
Adding a Cosmic SED method to the Pipeline

WillJRoper:
Since we may not always want to save spectra to calculate the Cosmic SED when running a ``Pipeline`` it seems worthwhile that we introduce a baked-in accumulation to collect the spectra as we go through the ``Pipeline``. 

This new method (and the observer frame equivalent) supports arbitrary galaxy attribute filters with an upper or lower bound (and the option to associate a custom label) to enable simple binning of the Cosmic SED. 

I have run this with the COLIBRE pipeline and got a bunch of Cosmic SEDs out comparing different things:
cosmic_sed_chabrier_los_sps_comparison_cosmic_sed_luminosity
cosmic_sed_chabrier_los_model_comparison_cosmic_sed_luminosity
cosmic_sed_chabrier_colibreLOS_intrinsic_attenuated_cosmic_sed_luminosity

And here using the binning functionality:
cosmic_sed_chabrier_colibreLOS_attenuated_by_mass_cosmic_sed_luminosity

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added cosmic SED aggregation across galaxy samples (rest-frame Lnu and observer-frame Fnu) with optional attribute filtering, averaging, volume normalization, labeled outputs, and writable outputs.

* **Documentation**
  * Added notebook example demonstrating cosmic SED and observed cosmic SED usage and integration into the workflow.

* **Tests**
  * Added and extended tests verifying cosmic SED sum/average behavior, volume normalization, error handling, and split/unsplit consistency.

* **Bug Fixes**
  * Minor docstring typo corrected.

### DIFF
--- a/src/synthesizer/pipeline/pipeline.py
+++ b/src/synthesizer/pipeline/pipeline.py
@@ -51,6 +51,7 @@ from synthesizer.pipeline.pipeline_utils import (
     get_full_memory,
     plot_timing_analysis,
     print_timing_analysis_table,
+    sanitise_hdf5_key_part,
     sum_dicts_recursive,
     validate_noise_unit_compatibility,
     write_timing_analysis_summary,
@@ -1838,11 +1839,11 @@ class Pipeline:
                         if op_kwargs["upper_bound"] is not None
                         else "inf"
                     )
-                    label = (
-                        f"{lower} < {op_kwargs['gal_attr']} < {upper}".replace(
-                            ".", "p"
-                        )
+                    label = sanitise_hdf5_key_part(
+                        f"{lower} < {op_kwargs['gal_attr']} < {upper}"
                     )
+            else:
+                label = sanitise_hdf5_key_part(label)
 
             # Loop over the spectra on the galaxy and all its components and
             # sum them into the cosmic SED dictionary on the pipeline with the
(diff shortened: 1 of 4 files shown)


WillJRoper:
This also now includes the ability to average the contributions to the cosmic SED and/or normalise it by an arbitrary volume. Both are specific to each call to `get_cosmic_sed` so you can set up any number of selections with or without averaging or voume normalisation.

src/synthesizer/pipeline/pipeline.py
```diff
@@ -1741,6 +1754,12 @@ def get_cosmic_sed(
                 provided and no bounds are provided the label will be "All".
                 If bounds are provided but no label is provided the label will
                 be "{lower_bound}<{gal_attr}<{upper_bound}". Default is None.
+            average (bool):
```
christopherlovell, line 1757: Why is this on the pipeline init? if it needs to be here, argument name very unclear

WillJRoper: I'm confused by your comment. This isn't on the init? It's an argument to the get_cosmic_Sed flagging method

christopherlovell: Apologies, I think I got confused by the github editor viewer

christopherlovell (approved):

PR #1131, opened 2026-04-21
Accounting for kernel overlaps in LOS column densities

WillJRoper:
This PR introduces the option to compute column densities using the overlap of the input (e.g. the stars) and source (e.g. the gas) kernels instead of considering the inputs as point sources.

This is a bit more involved since this method now has 3 regimes:
- kernels that are entirely behind the input kernel.
- kernels that overlap with the input kernel, but some of the overlap is behind the input.
- kernels that are entirely in front of the input kernel.

To handle these regimes, the ``Kernel`` object can now produce an overlap look-up table that explicitly encodes all three cases in terms of the projected kernel separation, the line-of-sight offset, and the ratio of the input and source smoothing lengths.

With this look-up table in hand, each input-source pair can be evaluated directly from a single table lookup and interpolation, rather than by sampling the input kernel during the runtime calculation.

This extra complexity does, of course, carry a performance penalty of around 5x.

Here is a comparison of the two methods. As you can see, the point approximation does a reasonable job but severely underpredicts the column density at low column densities and slightly overpredicts in dense environments.
LOS_comparison

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added smoothed-input LOS column-density mode (new as_points option) and overlap-kernel support with configurable build parameters; pipeline and particle APIs expose the option.

* **Documentation**
  * Fixed reStructuredText heading formatting.

* **Tests**
  * Expanded LOS test suite to cover smoothed-input modes, validation, consistency, and parameter effects.

* **Chores**
  * Reorganized kernel extension into modular components and added new example scripts demonstrating LOS computations and kernel geometry visualizations.


src/synthesizer/kernel_functions.py
```diff
@@ -22,6 +22,17 @@
...
+from synthesizer.extensions.kernel import compute_overlap_kernel
+from synthesizer.utils.operation_timers import timed, timer
+
+# Define the default overlap-kernel grid dimensions and bounds
+OVERLAP_Q_BINS = 64
+OVERLAP_U_BINS = 128
+OVERLAP_ETA_BINS = 48
+OVERLAP_ETA_MIN = 0.1
+OVERLAP_ETA_MAX = 10.0
+OVERLAP_BUILD_NDIM = 16
```
aswinpvijayan: Would be good to add comments on what these are individually

src/synthesizer/kernel_functions.py
```diff
@@ -85,6 +120,266 @@ def _integral_func(self, ii):
...
+        if self._truncated_los_kernel is not None:
+            bins = self._get_bins()
+            z_bins = np.linspace(-1.0, 1.0, 2 * self.binsize + 1)
+            return self._truncated_los_kernel.copy(), bins, z_bins
+
+        with timer("Kernel._get_truncated_los_kernel"):
+            # Get the projected-separation and LOS-coordinate bins and set up
+            # the output
+            bins = self._get_bins()
+            z_bins = np.linspace(-1.0, 1.0, 2 * self.binsize + 1)
```
aswinpvijayan: Don't you need the point '1.0' in there as well

src/synthesizer/kernel_functions.py
```diff
@@ -85,6 +120,266 @@ def _integral_func(self, ii):
...
+        Returns:
+            np.ndarray: The interpolated truncated-kernel values.
+        """
+        # Broadcast the projected-separation and LOS-coordinate inputs onto a
+        # common shape so the interpolation can be carried out element-wise.
+        q, z = np.broadcast_arrays(q, z)
+        values = np.zeros_like(q)
+
+        # Only points inside the projected kernel support can contribute.
+        valid = (q >= 0.0) & (q < 1.0)
```
aswinpvijayan: Shouldn't valid be (q >= 0.0) & (q<= 1.0)?

WillJRoper:
Converting this to a draft so we can implement some of the considerations here into the point version and then check the various kernel overlap regime integrals for correctness

src/synthesizer/kernel_functions.py
WillJRoper: These are now gone and are keyword arguments with descriptions.

src/synthesizer/kernel_functions.py
WillJRoper: linspace is inclusive upper bound so 1.0 is included.

src/synthesizer/kernel_functions.py
WillJRoper: This is all rewritten now, but we discussed this stuff at length 😅

WillJRoper:
We now have an example that demonstrates the geometry for the overlap and includes these plots. The former shows the geometry and colours the "input" kernel by the relative column density with respect to the whole kernel sight line for a slice through the centre of the kernel. The latter compares the resulting column density for all possible versions of the calculation.

overlap_geometry
overlap_comp

WillJRoper:
Given the growth of the kernel machinery here (and in preeceeding PRs), I have now split the kernel extensions into a set of submodules to make life easier moving forward.

christopherlovell (changes requested):
I'd prefer if the saved kernel was a hdf5 file or something safer than `.npz` (for some reason it won't let me comment directly on the relevant code). Happy to leave this for future though

WillJRoper:
I'll do that here. I had actually meant to make that switch 😅

WillJRoper:
Done

christopherlovell (approved):

PR #1136, opened 2026-05-01
Fix for PhotometryCollection unit issue

tHarvey303:
## Issue Type
- Bug

`PhotometryCollection - get_photo_fnu` appears to be storing fluxes as luminosities (.photo_lnu) not flux density (.photo_fnu) because the is_flux check doesn't work

For context it does this -`
is_flux = photometry.units == self.__class__.__dict__["photo_fnu"].unit
`
`photometry.units` is  `erg/(Hz*cm**2*s)` (due to the @accepts decorator) but `self.__class__.__dict__["photo_fnu"].unit`  is nJy. The == check doesn't check for dimensional equivalence, so despite nJy and `erg/(Hz*cm**2*s)`  being dimensionally equivalent the is_flux  evaluates to False, so the input photometry is stored as a luminosity and not a flux.

I've replaced it with this check (but split into two lines).

`is_flux = photometry.units.same_dimensions_as(self.__class__.__dict__["photo_fnu"].unit)`

I've also added a new test in the photometry test suite which tests that flux/luminosity units are preserved properly, which should catch issues like this in the future. 

## Checklist
- [x] I have read the [CONTRIBUTING.md]() -->
- [x] I have added docstrings to all methods
- [x] I have added sufficient comments to all lines
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no pep8 errors
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Release Notes

* **New Features**
  * Added correlated noise application for images and image collections via instrument noise maps.

* **Documentation**
  * Extended noise-application documentation with correlated noise examples and new example script demonstrating noise synthesis and visualization.

* **Improvements**
  * Enhanced unit handling in photometry calculations for improved consistency.

* **Tests**
  * Added comprehensive tests for correlated noise generation and photometry unit preservation.

PR #1140, opened 2026-05-07
(actually) fix PhotometryCollection unit issue

tHarvey303:
TLDR: All photometry is ~e32 too small at the moment -- see [here](https://synthesizer-project.github.io/synthesizer/auto_examples/parametric/plot_observed_sed.html) for an example.

The issue is in PhotometryCollection - #1136 fixed an issue with flux being stored as luminosity and vice-versa, but there was no unit conversion if we are passed a non-nJy flux unit. The C++ typically returns erg/(Hz*cm**2*s) and we strip the unit, whilst leaving the unit attached to the collection as nJy. This results in all the photometry being completely wrong.

I've switched to using unyt's inbuilt .to_value converter, but you may want to handle this slightly differently. 

## Summary by CodeRabbit

* **Refactor**
  * Internal handling of photometry data was changed to standardize unit conversion at construction; behavior for flux vs luminosity selection is unchanged and no user-facing API changes.

* **Tests**
  * Unit tests updated to compare numeric values after converting to expected units rather than asserting unit objects directly.

### DIFF
--- a/src/synthesizer/photometry.py
+++ b/src/synthesizer/photometry.py
@@ -85,7 +85,7 @@ class PhotometryCollection:
 
         # Keep raw ndarray storage and rely on Quantity descriptors for
         # units.
-        self._photometry_data = photometry.ndview
+        self._photometry_data = photometry.to_value(fnu_unit)
 
         if is_flux:
             self.photo_fnu = self._photometry_data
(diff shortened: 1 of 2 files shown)


WillJRoper (changes requested):
Sorry to be a stickler, but can you use `unyt_to_ndview`? This is our version of `to_value` (effectively), but it a) doesn't make a copy and b) does the conversion in place. Since this is photometry, the hidden allocation may not be *too costly* but since we have this machinery already it's worth using.

WillJRoper (approved):

Issue #1147, opened 2026-05-26
Allow a parametric Stars object to be defined by its surviving mass instead of initial mass


stephenmwilkins:
Include the functionality to define a stellar population (`Stars` object) by its surviving mass instead of its initial mass.

In many use cases a user may want to create a stellar population with a known current mass, or surviving mass.

PR #1151, opened 2026-05-26
Define by surviving mass

stephenmwilkins:
Addresses #1147 #1152 

- It is now possible to define the surviving mass when building a parametric stellar population (`Stars` object).
- For clarity all of the normalisation has been moved to the end and the key normalisation factor is retained for further use. This will become useful when for example you want to regenerate the star formation history at an earlier time.
- Caught a bug where things broke if `initial_mass` was not provided.
- Added examples of the `surviving_mass` functionality.

## Issue Type
- [x] Bug
- [x] Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Stars component now accepts surviving stellar mass as an alternative to initial stellar mass for parameterizing galaxies, and ensures consistent mass normalization across different SFH/metallicity input methods.

* **Documentation**
  * Updated notebook with examples demonstrating surviving stellar mass usage, required setup (grid with stellar fraction), instantiation variations, and corresponding explanatory plots.


christopherlovell:
Don't we need to include the `surviving_mass` attributes with defaults on a grid object? How are these interpolated for arbitrary SFHs, using the same grid interpolation technique?

stephenmwilkins:
> Don't we need to include the `surviving_mass` attributes with defaults on a grid object? How are these interpolated for arbitrary SFHs, using the same grid interpolation technique?

I'm not sure I understand? The key feature from a grid is the `stellar_fraction` which encodes the fraction of stars remaining as a function of age/metallicity. It might be more intuitive to rename this `surviving_fraction` though.

src/synthesizer/parametric/stars.py
```diff
@@ -237,6 +263,36 @@ def __init__(
...
+                "provided to get the stellar_fraction for rescaling the SFZH!"
+            )
+
+        # Get the stellar fraction from the grid if we have been given a
+        # surviving mass, this is needed to rescale the SFZH to obey the
+        # surviving mass constraint. This is the fraction of the initial mass
+        # that survives to the present day, and is used to calculate the
+        # initial mass from the surviving mass.
+        if self.surviving_mass is not None and grid is not None:
+            self.stellar_fraction = grid.stellar_fraction
```
christopherlovell, line 297: Right, I meant 'stellar_fraction`. Where is this defined on a grid? Should we include some defaults? And same question as above, how is this interpolated? 

Right now this line will error if it's not present on the grid object.

christopherlovell (changes requested):

src/synthesizer/parametric/stars.py
WillJRoper: I think this is a "you needa grid with it" situation. The function should check the grid has the required attribute and if not, throw an error. Hopefully, as time goes on we will add it to all the grids it can be added to, but there may always be certain models we can't include it for.

src/synthesizer/parametric/stars.py
```diff
@@ -97,20 +105,25 @@ class Stars(StarsComponent):
 
     # Define quantities
     initial_mass = Quantity("mass")
+    surviving_mass = Quantity("mass")
 
     @accepts(initial_mass=Msun.in_base("galactic"))
+    @accepts(surviving_mass=Msun.in_base("galactic"))
```
WillJRoper: `accepts` takes a list of arguments, calliing it twice just doubles the overhead.

src/synthesizer/parametric/stars.py
```diff
@@ -97,20 +105,25 @@ class Stars(StarsComponent):
...
         initial_mass=None,
+        surviving_mass=None,
+        grid=None,
         morphology=None,
         sfzh=None,
         sf_hist=None,
         metal_dist=None,
         fesc=None,
         fesc_ly_alpha=None,
+        age_offset=0 * yr,
```
WillJRoper: age offset should in accepts too

src/synthesizer/parametric/stars.py
```diff
@@ -159,6 +179,12 @@ def __init__(
                 The escape fraction of incident radiation from the stars.
             fesc_ly_alpha (float):
                 The escape fraction of Ly-alpha radiation from the stars.
+            age_offset (unyt_quantity/float):
```
WillJRoper: Rather than an argument to the init, wouldn't this be better done as a method which returns a new Stars object with the modification? Feels like this could end up messy as implemented

src/synthesizer/parametric/stars.py
```diff
@@ -256,6 +312,7 @@ def __init__(
             else:
                 # Otherwise calculate the total initial mass
                 self._initial_mass = np.sum(self.sfzh)
+                self.initial_mass = self._initial_mass * Msun
```
WillJRoper: This isn't required? The Quantity handles this automatically? You don't need this and the line above.

src/synthesizer/parametric/stars.py
```diff
@@ -339,29 +396,48 @@ def _get_sfzh(self, instant_sf, instant_metallicity):
         # If both are instantaneous then we can do the whole SFZH in one go
         if instant_sf is not None and instant_metallicity is not None:
             inst_stars = ParticleStars(
-                initial_masses=np.array([self._initial_mass]) * Msun,
-                ages=np.array([instant_sf.to("yr").value]) * yr,
+                initial_masses=np.array([1.0]) * Msun,
```
WillJRoper, line 406: Could be worth a comment explaining why we use 1.0 here because it looks confusing when initial_mass has been passed.

src/synthesizer/parametric/stars.py
```diff
@@ -339,29 +396,48 @@ def _get_sfzh(self, instant_sf, instant_metallicity):
...
+                ages=np.array(
+                    [
+                        instant_sf.to("yr").value
+                        + self.age_offset.to("yr").value
```
WillJRoper: I really don't like this as is 😅 as stated above this should be a transformation done in a method.

WillJRoper: Also the to is not required if `accepts` is used.

src/synthesizer/parametric/stars.py
```diff
@@ -339,29 +396,48 @@ def _get_sfzh(self, instant_sf, instant_metallicity):
...
+            # a normalisation based on the initial mass or surviving mass, so
+            # we need to continue with the rest of the code to do that if
+            # necessary
 
         # Handle the instantaneous SFH case
         elif instant_sf is not None and instant_metallicity is None:
             inst_stars = ParticleStars(
-                initial_masses=np.array([self._initial_mass]) * Msun,
-                ages=np.array([instant_sf.to("yr").value]) * yr,
+                initial_masses=np.array([1.0]) * Msun,
```
WillJRoper, line 438: Same comment as above.

WillJRoper: And all the other places

WillJRoper (changes requested):

src/synthesizer/parametric/stars.py
stephenmwilkins: Done

src/synthesizer/parametric/stars.py
stephenmwilkins: Done!

src/synthesizer/parametric/stars.py
stephenmwilkins: I had wondered that, but doesn't that then effectively you end up making two stars objects, one with and without the offset?

In reality that is how `synthpop` actually has to work so maybe it's fine?

Happy to try and implement that if you think it's the best approach.

src/synthesizer/parametric/stars.py
stephenmwilkins: Done

src/synthesizer/parametric/stars.py
stephenmwilkins: Done

src/synthesizer/parametric/stars.py
stephenmwilkins: I completely understand that. The problem though is I think the other way to do it requires essentially replicating everything in `_get_sfzh()` which seems bad from a maintenance point of view.

WillJRoper: In this case that is the lesser of two evils IMO. There's probably common functionality that can be refactored out of both at a later date anyway.

src/synthesizer/parametric/stars.py
stephenmwilkins: I've added a check for this with exception.

stephenmwilkins:
I think the requested changes have now been inplemented.

christopherlovell (approved):

WillJRoper (approved):

PR #1160, opened 2026-05-29
Added method to expicitly overload attributes and set fixed parameters on EmissionModel

WillJRoper:
Recent conversations about overriding grid axis attributes got me thinking. The machinery we have 100% supports remapping and fixing parameters, but it is not as obvious as it should be and isn't very ergonomic when working with premade models. 

To this end, I have added 2 new methods, `set_fixed_parameter` and `set_attribute_overload`, which allow you to do this at any time and make the process a lot safer and, importantly, ergonomic.

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Expanded docs on overriding emission model parameters with clear examples for setting, redirecting, and clearing fixed parameters after instantiation.

* **New Features**
  * Added runtime controls to set, redirect, and remove fixed parameters and attribute lookups on emission models, allowing dynamic adjustment of parameter sources and overrides.


src/synthesizer/emission_models/base_model.py
```diff
@@ -830,6 +830,90 @@ def _set_attr(self, attr, value):
...
+        When get_param is called to access a parameter, it will first check
+        the fixed_parameters dictionary on the model, prior to looking for
+        the parameter on the emitter.
+
+        Args:
+            param_name (str): The name of the parameter to fix.
+            value (Any): The value to fix the parameter to.
+        """
+        # Ensure we don't already have an override for this parameter
+        if param_name in self.fixed_parameters:
```
christopherlovell, line 892: Could just overwrite?

christopherlovell (approved):
Optional change above, otherwise good to go

src/synthesizer/emission_models/base_model.py
WillJRoper: My worry with silently overwriting is that you could end up confused. I couldn't think of a time when you'd rewrite the fixed parameters over and over so I leant towards the safety of explicitly rejecting

PR #1168, opened 2026-06-03
Mammoth "adaptive extension precision" PR

WillJRoper:
This PR contains a wide-reaching set of changes enabling variable floating precision in the C++ extensions. Previously, we have been pinned to double precision, which is OTT in many applications and leads to large memory footprints when we don't need them. 

I have tested multiple ways to do this, but they were too rigid or complex. Here I have settled on leveraging C++ templating so that the extensions actually adapt to the precision of the inputs. This isn't 100% flexible because certain "classes" of input all must share the same precision (i.e. all particle properties or all grid properties etc.). I have propagated an output precision to the user interface, though. On all methods that interface with C++, the user can now pass a `out_dtype=np.float32/64` which will cause the output (i.e. spectra, photometry, image etc.) to be computed in that precision. This now defaults to float32 except for blackhole spectra and bolometric luminosities, which readily overflow and instead default to float64. 

This PR will stay in draft, and I will split small chunks of it out into smaller PRs for the ease of reviewing. I will use this as a central record, though, sharing plots etc.

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Global configurable output precision (out_dtype) for spectra, photometry, integration, lines, images, and pipeline; Grid can convert/initialize precisions.
  * Multiple profiling/benchmarking scripts for performance and memory scaling.

* **Improvements**
  * Broad mixed‑precision support across numeric codepaths; photometry/filters and integration honor requested output dtype.
  * CI now checks C++ formatting; added a formatting script.

* **Chores**
  * Bumped ruff dev tool version.

* **Tests**
  * Added/updated tests validating precision behavior and new APIs.


src/synthesizer/emission_models/transformers/dust_attenuation.py
```diff
@@ -1584,6 +1590,7 @@ def get_tau_at_lam(
                 valid,
                 None,
                 False,
+                np.float32,
```
aswinpvijayan, line 2066: Should we change this to reflect the new template? I just defined the array as np.float32 in there, without much reason other than smaller memory footprint and since it doesn't need that much precision.

src/synthesizer/emissions/sed.py
```diff
@@ -652,6 +661,7 @@ def bolometric_luminosity(self):
             self._nu,
             self._lnu,
             method="trapz",
+            out_dtype=np.float64,
```
aswinpvijayan, line 664: Is this because the bolometric luminosity would naturally overflow?

WillJRoper: Exactly, it's a royal pain. We could change the unit system we use because of it, though... as a temporary solution, this was a small sacrifice.

src/synthesizer/emission_models/transformers/dust_attenuation.py
WillJRoper: What do you mean exactly? If this is already float32, isn't that fine? (I may be missing something obvious)

aswinpvijayan: I was just wondering if we wanted to keep the same templating as the rest of this dtype change. It is fine as it is.

WillJRoper: Ah, fair. I'm honestly ambivalent. It won't save that much memory, but is probably ok as is anyway.

docs/source/performance/precision.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Understanding the errors
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Because Synthesizer refuses to cast inputs behind your back, you may see
+errors like:
+
+.. code-block:: text
+
+    TypeError: ages must share the same floating-point dtype as masses
```
christopherlovell, line 130: Could add this to the FAQ in anticipation...

src/synthesizer/emissions/line.py
```diff
@@ -1357,9 +1394,16 @@ def apply_attenuation(
                 cont=get_array_quantity_view(att_cont, cont_units),
             )
 
-        # Compute the transmission for the remaining generic cases.
-        transmission = dust_curve.get_transmission(
-            tau_v, self.lam, **dust_curve_kwargs
+        # Compute the transmission for the remaining generic cases. The curve
+        # is evaluated at the emission dtype (with overflow trapped) so the
+        # transmission is born at the right precision rather than computed at
```
christopherlovell: the word born is a bit odd here, initialised?

src/synthesizer/emissions/sed.py
```diff
@@ -133,8 +134,16 @@ def __init__(self, lam, lnu=None, description=None):
         # Set the wavelength
         self.lam = lam
 
-        # Calculate frequency
-        self.nu = c / self.lam
+        # Calculate frequency. Dividing by the `c` physical-constant
```
christopherlovell: does this implicitly copy? maybe we define our own constants?

src/synthesizer/load_data/__init__.py
```diff
@@ -0,0 +1,32 @@
+"""Load simulation particle data into Synthesizer galaxy objects.
```
christopherlovell: not necessatrily just particle data, this contains SAM loaders too

src/synthesizer/load_data/load_yt.py
```diff
@@ -1705,6 +1705,30 @@ def _infer_centre(stars, gas, black_holes):
     return None
 
 
+def _cast_component_dtype(component_dict, dtype):
```
christopherlovell: could this be generalised for components from other sources? Are there situations where certain components should be in different dtypes?

src/synthesizer/utils/precision.py
```diff
@@ -0,0 +1,95 @@
...
+Note that ``out_dtype`` only controls output arrays. Input arrays (particle
+data and grids) are used at whatever precision they are provided at and are
+never cast or copied behind the scenes; see the precision documentation for
+the rules on mixing input precisions.
+"""
+
+import numpy as np
+
+# The allowed floating point dtypes for outputs.
+_ALLOWED_DTYPES = (np.dtype(np.float32), np.dtype(np.float64))
```
christopherlovell, line 28: are these really the only allowed? could be interesting to do some tests for photometry with 16 point fp to see how much you save

tests/data/create_golden_pipeline_reference.py
```diff
@@ -0,0 +1,60 @@
+"""Regenerate the golden full-pipeline regression reference.
+
+This script computes the reference output used by
+``tests/test_pipeline_regression.py`` and writes it to
+``golden_pipeline_reference.npz``.
```
christopherlovell: can we switch this to hdf5?

christopherlovell (approved):

src/synthesizer/emissions/sed.py
WillJRoper: Yes, that actually will make a copy. I've swapped it out for an in-place multiplication to avoid the various casts and conversions.

I had been wondering about consolidating a bunch of our constants into a constants module for convenience. Will raise an issue. It could also provide helpers for this sort of thing.

src/synthesizer/utils/precision.py
WillJRoper: Let's test this in the future, it'll require adding all the templates in the backend too.

PR #1175, opened 2026-06-07
Create at earlier age

stephenmwilkins:
Adds a method to `Stars` which returns a new `Stars` object at an earlier time. The necessitated some significant refactoring and simplification.

Also added an example.

## Issue Type
- [x] Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Stars can now be initialized using `surviving_mass` as an alternative to `initial_mass`.
  * Added `get_at_earlier_time()` method to retrieve a Stars instance at an earlier age.

* **Documentation**
  * Updated notebook with example demonstrating parametric Stars creation and time-based remapping.


src/synthesizer/parametric/stars.py
```diff
@@ -215,19 +237,19 @@ def __init__(
...
         elif isinstance(metal_dist, (unyt_quantity, float, np.floating)):
-            instant_metallicity = metal_dist  # an instantaneous SFH
+            self.instant_metallicity = metal_dist  # an instantaneous SFH
             self.metal_dist_func = None
             self.metal_dist = None
         elif isinstance(metal_dist, (unyt_array, np.ndarray)):
             self.metal_dist = metal_dist  # a numpy array
             self.metal_dist_func = None
-            instant_metallicity = None
+            self.instant_metallicity = None
```
WillJRoper: The reason these weren't attributes originally is that they are an internally useful variable, not externally. If you think they really need to be attributes, then they should at least be private.

src/synthesizer/parametric/stars.py
```diff
@@ -303,108 +370,155 @@ def __init__(
             # Irregular
             self.metallicity_grid_type = None
 
-    @accepts(instant_sf=yr)
-    @timed("ParametricStars._get_sfzh")
-    def _get_sfzh(self, instant_sf, instant_metallicity):
+    # @accepts(age_offset=yr.in_base("galactic"))
+    @timed("ParametricStars._create_sfzh")
+    def _create_sfzh(self, age_offset=0 * yr, reset_sf_hist_metal_dist=False):
```
WillJRoper: use `_get_` nomenclature. It's important that elsewhere in the code these things match.

src/synthesizer/parametric/stars.py
```diff
@@ -303,108 +370,155 @@ def __init__(
...
         """Compute the SFZH for all possible combinations of input.
 
         If functions are passed for sf_hist_func and metal_dist_func then
         the SFH and ZH arrays are computed first.
 
         Args:
-            instant_sf (unyt_quantity/float):
```
WillJRoper, line 375: Still prefer these being arguments

src/synthesizer/parametric/stars.py
```diff
@@ -303,108 +370,155 @@ def __init__(
...
-                An age at which to compute an instantaneous SFH, i.e. all
-                stellar mass populating a single SFH bin. Note, this must
-                be the age itself, not the log10(age).
-            instant_metallicity (float):
-                A metallicity at which to compute an instantaneous ZH, i.e. all
-                stellar populating a single ZH bin. Note, this must be the
-                metallicity itself, not the log10(metallicity).
+            age_offset (unyt_quantity):
+                The offset to apply to the age grid when calculating the SFZH
+            reset_sf_hist_metal_dist:
```
WillJRoper: Is there ever going to be a case where age_offset is not passed and this is True? From this description, it seems the argument is redundant since we use it when calculating SFZH at an earlier time, which necessarily means age_offset is in use.

src/synthesizer/parametric/stars.py
```diff
@@ -418,39 +532,50 @@ def _get_sfzh(self, instant_sf, instant_metallicity):
...
             )
 
         # Finally, calculate the SFZH grid based on the above calculations
-        self.sfzh = self.sf_hist[:, np.newaxis] * self.metal_dist
+        return sf_hist[:, np.newaxis] * metal_dist
 
-        # Normalise the SFZH grid if needs be
-        if self.initial_mass is not None:
-            self.sfzh /= np.sum(self.sfzh)
+    # @accepts(age_offset=yr.in_base("galactic"))
```
WillJRoper: Why is this commented?

WillJRoper (changes requested):
A few initial things. Need to look a bit deeper once addressed. What is this waiting for? I assume it contains the other branch's changes too?

src/synthesizer/parametric/stars.py
stephenmwilkins: If I recall correctly it's because the variable is needed again if the `get_stars_at_earlier` times method is called. I suppose it can then become private, though I think it being an attribute also makes sense. After all, we keep knowledge of the SFH function in that case.

stephenmwilkins:
Yes, waiting for the other PR

src/synthesizer/parametric/stars.py
stephenmwilkins: That's true. I suppose by default `age_offset` could be `None` in which case the reset happens and age_offset is set to 0.0? Would that be better?

src/synthesizer/parametric/stars.py
WillJRoper: If it needs holding on to, that's fine (though I don't see why it is needed; you always have the sf and metallicity arrays so finding the same arrays at an earlier times should not require the input values). 

My issue is that the user doesn't need it, so it should be private. Partly so it isn't included in object printouts.

stephenmwilkins: You can in principle reverse engineer the input values but it's messy. You can't simply shift things because of the logarithmic spacing. 

I've made it private though.

src/synthesizer/parametric/stars.py
stephenmwilkins: Done

src/synthesizer/parametric/stars.py
stephenmwilkins: Done

src/synthesizer/parametric/stars.py
stephenmwilkins: Fixed

docs/source/galaxy_components/stars.ipynb
```diff
@@ -459,11 +459,11 @@
    "source": [
     "# Define a delta function for metallicity\n",
     "metal_dist = ZDist.DeltaConstant(log10metallicity=-2.5)\n",
-    "print(metal_dist)\n",
+    "# print(metal_dist)\n",
```
WillJRoper: Why is there commented out code in the docs?

docs/source/galaxy_components/stars.ipynb
```diff
@@ -459,11 +459,11 @@
...
     "\n",
     "# Define a constant SFH\n",
     "sfh = SFH.Constant(100 * Myr)\n",
-    "print(sfh)\n",
+    "# print(sfh)\n",
```
WillJRoper: Same

WillJRoper (changes requested):
Most minor of comments but otherwise looking good.

docs/source/galaxy_components/stars.ipynb
stephenmwilkins: No idea why I commented them out. Gone now.

stephenmwilkins:
Anyone understand the error?

WillJRoper (approved):

PR #1177, opened 2026-06-15
Force float64 type on SFH arrays to avoid precision errors.

tHarvey303:
## Issue Type
- Bug

I found an edge case in the parametric SFH history case where the age array generation with `np.arange` (in `SFH.calculate_sfh`) genetates an np.int64 array. The use of `np.zeros_like` in the vectorized `_sfrs` method then propagates this into the `sfrs` array, which then silently clips SFH values to integers.

Example 
```
from unyt import Gyr
from synthesizer.parametric import SFH

sfh = SFH.LogNormal(tau=0.9, peak_age=1.1*Gyr, max_age=3*Gyr)
sfh.plot_sfh()
```

image

With this fix (forcing the type to np.float64) we get the correct output. We may want to actually fix this differently but regardless it should be a simple fix to make (before we replace all the parametric machinery anyway.

image

## Checklist


WillJRoper (approved):
How fun 😅

PR #1178, opened 2026-06-16
Fix redshift dtype issue in flux density calculation

tHarvey303:
Fix an edge case I encountered in synference where if your redshift has a numpy dtype and is not a python float, the `compute_fnu` C extension outputs perfect 0's for flux density rather than the correct value.

This simply casts the redshift to a float before it is used to ensure it works. There could be other versions of this edge case in other C extensions as well

## Issue Type
- Bug

## Checklist

## Release Notes

* **Bug Fixes**
  * Improved numeric precision in redshift-related calculations for flux computations.


WillJRoper (approved):
For now we need this but it should be removed in the near future as we merge the adaptive precision machinery

PR #1182, opened 2026-06-18
Implementing simple spatial resampling of Particle objects

### LINKED ISSUE #891: Allow resampling of SFH and metallicity functions
At the moment we have a function that resamples a `SFZH` array. However, it would be nice to also have a function that sampled a SFH and metallicity function instead. 

This could be achieved either in the new function or through a e.g. `sample` method included in the SFH/Z classes themselves (or both, i.e. with a wrapper that seamlessly produces a new particle.stars object).

### LINKED ISSUE #892: Improved sample_sfzh
`sample_sfzh` could be improved in two ways:

- It only returns ages at exactly the grid point. Ideally we'd sample the grid point bin instead to give a continuous range of ages.
- It doesn't take account of the total mass in the SFZH array, instead relying on the user to provide the initial masses. A simple fix here would be to force the sum of the resampled particle masses to be equal to the total mass in the SFZH array.

It might be also good to rename to `sample_sfzh_from_array` to avoid confusion with a function that samples directly from functions as suggested in #891.

### LINKED ISSUE #1071: Spatially resample a star particle into individual stars
It would - eventually - be useful to be able to sample star particles into individual stars. This would simply take the IMF as an argument and the sampling approach.

The real utility of this though would be assigning spectra to individual stars but this would require quite a different type of grid.

WillJRoper:
This PR introduces methods and machinery to randomly resample any particle object across SPH kernels into N new particles with or without masks. Attributes can be resampled in a number of different ways:

- "duplicated" — Every child inherits the parent value unchanged. Default for intensive quantities (metallicity, optical depth, alpha enhancement, elemental abundances, custom attributes).
- "proportional" — The parent value is divided equally among children (value / resample_factor), conserving the total sum. Default for extensive quantities (initial_mass, current_mass, masses, dust_masses).
- "normal" — Each child is drawn from Normal(value, σ) independently. σ defaults to the population standard deviation; can be fixed via ("normal", sigma).
- "lognormal" — Like normal but in log-space (LogNormal(log(value), σ_log)), keeping values positive. Best for metallicities, SFR, ages.
- "conserved_normal" — Starts from the proportional split, adds Gaussian scatter, then renormalizes each group so children sum exactly to the parent value. Scatter with exact conservation.
- SFH/MetalDist sampling — When a parametric SFH or metallicity distribution is provided to Stars.spatially_resample, ages and metallicities are continuously sampled from 1‑D or 2‑D (SFZH) distributions instead of the mode-based resampling, with masses conserved to the original total.

This lays the groundwork for more refined resampling techniques in the future that better respect the matter field.

Closes #1071. Closes #892. Closes #891.

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added spatial resampling for gas and stars with modes (`duplicated`, `proportional`, `normal`, `lognormal`, `conserved_normal`), optional velocity dispersion, and mask-based selective resampling.
  * Added parametric SFH/metallicity (SFZH) resampling for stellar ages and metallicities.
  * Added a clear error for unsupported black hole spatial resampling.
* **Documentation**
  * Added new particle spatial-resampling example scripts (resampling, distributions, metallicity, SFH).
* **Bug Fixes**
  * Improved unit handling when extracting emission-model parameters/axes.
  * Fixed integration calls by passing the expected trailing argument.
* **Tests**
  * Added comprehensive spatial-resampling test coverage for utilities and gas/stars APIs.


src/synthesizer/particle/resample_utils.py
```diff
@@ -0,0 +1,678 @@
...
+        )
+    return new_vels
+
+
+def resample_smoothing_lengths(smoothing_lengths, resample_factor):
+    """Scale smoothing lengths for volume conservation and tile.
+
+    Each smoothing length is scaled by ``resample_factor ** (-1/3)`` so that
+    the total kernel volume is conserved across the resampled particles, then
+    tiled *resample_factor* times.
```
aswinpvijayan: Would be good to say explicitly that the returned smoothing lengths are the same

src/synthesizer/particle/resample_utils.py
```diff
@@ -0,0 +1,678 @@
...
+        unyt_array, (N * resample_factor,):
+            Resampled smoothing lengths.
+    """
+    return _tile_array(
+        smoothing_lengths / (resample_factor ** (1.0 / 3.0)),
+        resample_factor,
+    )
+
+
+def _sample_sfzh_arrays(sfzh, log10ages, metallicities, nstar, rng):
```
aswinpvijayan, line 510: I understand that this is all internal, but any reason why input age is in log10 while output is non-logged?

WillJRoper: This just boils down to the fact that the `Grid` is in terms of log10 age, meaning most SFZH histories adopt that same convention. Means the spacing on that access for the SFZH is a bit smoother too.

src/synthesizer/particle/resample_utils.py
WillJRoper: They aren't the same, they're scaled to conserve the total volume. I have added a sentence to make that clear.

src/synthesizer/particle/stars.py
```diff
@@ -814,6 +903,552 @@ def renormalise_mass(self, stellar_mass):
...
+
+        Per-attribute resampling is controlled via *attr_modes*.  See
+        :func:`~.resample_utils.resample_by_mode` for available modes.
+
+        **Default modes per attribute:**
+
+        - ``initial_masses``, ``current_masses`` → ``"proportional"``
+        - ``ages``, ``metallicities``, ``alpha_enhancement``, ``s_oxygen``,
+          ``s_hydrogen`` → ``"duplicated"``
+        - ``tau_v``, ``softening_lengths`` → ``"duplicated"`` if per-particle,
```
aswinpvijayan: I guess if its softening, it should remain the same, unless one of your variable softening length sims. But quantities like tau should be recalculated if the coordinates are being shifted. Might be a good idea to put warning for quantities that need to be recalculated after sampling.

src/synthesizer/particle/stars.py
```diff
@@ -814,6 +903,552 @@ def renormalise_mass(self, stellar_mass):
...
+                    attr_modes.get("current_masses", "proportional"),
+                    resample_factor,
+                    rng,
+                )
+                if current_masses is not None
+                else None
+            )
+
+        # Optical depth only needs resampling if it's an array of values for
+        # each particle, if it's a single value we just keep it as-is
```
aswinpvijayan, line 1279: I think this should be recalculated instead of resampled

aswinpvijayan (changes requested):

src/synthesizer/particle/resample_utils.py
aswinpvijayan: I meant its the same smoothing length for each sub-sample.

src/synthesizer/particle/stars.py
WillJRoper: That's down to the user, really. Not something that should be automatically handled, given how many different ways it could be interpreted. Maybe best to just document that clearly. 

When using this, I always reran the LOS calculation, but if running with a fixed screen per particle or in general, then hanging on to the value is probably best. I will make that the case and add some documentation flagging this.

src/synthesizer/particle/stars.py
WillJRoper: AS with the comment below.

WillJRoper:
Dealt with the comments (mainly adding documentation to be clear about things). One meaningful change: softenings are now actively resampled because they would be different if the resolution increases. Of course, in reality I can't think of a single time we would ever have the softenings actually attached to an object... but they are there and handled properly should that day arrive 😂

aswinpvijayan (approved):

PR #1184, opened 2026-06-20
Add Redshift Space Distortion (RSD) effects

christopherlovell:
Adds RSD effects from peculiar velocities at the galaxy level. Importantly it passes this at the `get_fnu` level; rest frame spectra are generated once, and RSD effects applied afterwards. Luminosity distances and IGM effects assume the cosmological redshift `z`, only the wavelength redshift includes the peculiar velocity contribution.

This functionality is needed for forward-modelling lightcones, where a galaxy's observed flux carries its bulk motion relative to the Hubble flow on top of the cosmological redshift. The change is fully backward compatible.

## What changed

- **`Sed.get_fnu(cosmo, z, igm=None, nthreads=1, peculiar_velocity=None)`**
  When `peculiar_velocity` is given, `z` is treated as the *cosmological*
  redshift and the spectrum is shifted to the observed redshift `z_obs`, with
  `1 + z_obs = (1 + z)(1 + v_los / c)`:
  - the wavelength shift and the `(1 + z)` flux factor use `z_obs`;
  - the luminosity distance uses the true `z` (effective
    `d_L = d_L(z)·(1 + z_obs)/(1 + z)`), since a peculiar velocity is a motion,
    not a displacement, and does not change the source distance;
  - the IGM attenuation uses the true `z`.
  Accepts a unyt velocity or a float interpreted as km/s; positive = receding.

- **`Galaxy.get_observed_spectra(..., peculiar_velocity=None)`**
  Passes `peculiar_velocity` through to every `Sed.get_fnu` call (integrated
  and per-particle), so the bulk shift is applied consistently. An explicit
  argument overrides a `peculiar_velocity` attribute set on the galaxy; if both
  are absent the behaviour is unchanged.

## Issue Type
- Enhancement

## Checklist

## New Features
* Observed spectrum calculations now accept an optional line-of-sight peculiar velocity to improve observer-frame redshift accuracy and spectral conversions.
* Spectrum generation (including galaxy, stellar, particle, and black hole contributions) now applies the velocity-dependent observed redshift consistently when provided.

## Tests
* Added coverage for `peculiar_velocity` behavior: `None`/zero matches default cosmological results, while nonzero values produce the expected observed-frame redshift shift.

### DIFF
--- a/src/synthesizer/emissions/sed.py
+++ b/src/synthesizer/emissions/sed.py
@@ -1130,7 +1130,7 @@ class Sed:
         return get_quantity_view(self, "_fnu")
 
     @timed("Sed.get_fnu")
-    def get_fnu(self, cosmo, z, igm=None, nthreads=1):
+    def get_fnu(self, cosmo, z, igm=None, nthreads=1, peculiar_velocity=None):
         """Calculate the observed frame spectral energy distribution.
 
         This will also populate the observed wavelength and frequency arrays
@@ -1144,12 +1144,18 @@ class Sed:
             cosmo (astropy.cosmology):
                 astropy cosmology instance.
             z (float):
-                The redshift of the spectra.
+                The cosmological redshift of the spectra (sets luminosity
+                distance and IGM).
             igm (igm):
                 The IGM class. e.g. `synthesizer.igm.Inoue14`.
                 Defaults to None.
             nthreads (int):
                 The number of threads to use for the bulk flux conversion.
+            peculiar_velocity (unyt_quantity/float):
+                Line-of-sight peculiar velocity (positive = receding), a unyt
+                velocity or a float in km/s. Shifts the spectrum to z_obs with
+                1 + z_obs = (1 + z)(1 + v / c); distance and IGM stay at z.
+                Defaults to None (no shift).
 
         Returns:
             fnu (ndarray)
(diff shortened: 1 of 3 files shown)


WillJRoper:
More than happy for this go in, but for a particle galaxy shouldn't you be using the self-consistent velocity shifting by setting `vel_shift=True` on the `EmissionModel`?

src/synthesizer/emissions/sed.py
```diff
@@ -1144,12 +1144,18 @@ def get_fnu(self, cosmo, z, igm=None, nthreads=1):
...
             z (float):
-                The redshift of the spectra.
+                The cosmological redshift of the spectra (sets luminosity
+                distance and IGM).
             igm (igm):
                 The IGM class. e.g. `synthesizer.igm.Inoue14`.
                 Defaults to None.
             nthreads (int):
                 The number of threads to use for the bulk flux conversion.
+            peculiar_velocity (unyt_quantity/float):
```
WillJRoper, line 1154: Can we not accept float and instead use the accepts decorator to guarantee?

src/synthesizer/emissions/sed.py
```diff
@@ -1164,6 +1170,17 @@ def get_fnu(self, cosmo, z, igm=None, nthreads=1):
         if self.redshift == 0:
             return self.get_fnu0()
 
+        # Peculiar velocity shifts the spectrum to z_obs without moving the
+        # source; z_obs == z when no velocity is given.
+        if peculiar_velocity is not None:
+            if hasattr(peculiar_velocity, "units"):
```
WillJRoper, line 1176: This is much cleaner if we use accepts too

WillJRoper (changes requested):

christopherlovell:
> More than happy for this go in, but for a particle galaxy shouldn't you be using the self-consistent velocity shifting by setting `vel_shift=True` on the `EmissionModel`?

As discussed offline, this is subtly different for performance reasons, but we should definitely simplify the documentation to clarify the overlap and any differences between these methods (and ensure they can't be double applied)

Issue #1183, opened 2026-06-18
Negative pixels introduced in resampling


madelinemarshall:
When convolving a single point source of flux with some JWST NIRCam PSFs from STPSF (F115W and F200W for example), the resulting image has negative pixels. The STPSFs are sampled on pixels that are 4x finer than my desired Synthesizer image pixel scale. This error seems to be because of the spline interpolation that deals with the resampling, which overshoots/undershoots around sharp features and isn't flux preserving.

As a fix, I have made a simple binning / uniform splitting method that works when you are resampling by an integer factor. This conserves flux across pixel scales. However, I'm guessing that the uniform splitting method may not be ideal for some scenarios. For the simple case of a highly oversampled PSF model from STPSF, convolving with the science image and reducing to the desired NIRCam pixel scale, I think this works well, and does fix the negative pixel issue. Not sure if you want to implement this in the full distributed version, but wanted to flag this issue as it will be particularly relevant for point sources (e.g. AGN images).

Core of JWST NIRCam F115W PSF, produced by Synthesizer imaging a single point source. Black pixels = negative flux
<img width="767" height="662" alt="Image" src="https://github.com/user-attachments/assets/d735778e-cad3-4220-a5d8-2c313c8dcf95" />
No negative pixels with new implementation.
<img width="729" height="571" alt="Image" src="https://github.com/user-attachments/assets/2904e192-b343-428c-894f-2c4ec8c6becb" />

Here's the updated code I use for this fix.
```python

def resample(self, factor):
        """Resample the image by factor.

        Args:
            factor (float):
                The factor by which to resample the image, >1 increases
                resolution, <1 decreases resolution.
        """
        # Resample the resoltion, this will also update the npix and fov (if
        # necessary)
        self._resample_resolution(factor)

        # Resample the image.
        # NOTE: skimage.transform.pyramid_gaussian is more efficient but adds
        #       another dependency.
        if self.arr is not None:
            int_factor = int(round(factor))
            inv_factor = int(round(1.0 / factor))

            if np.isclose(factor, int_factor):
                # Replace each pix with an int_factor × int_factor block whose subpixels each contain 1/int_factor**2
                # of the original pixel value, preserving total flux
                self.arr = np.repeat(np.repeat(self.arr, int_factor, axis=0),int_factor,axis=1) / int_factor**2

            elif np.isclose(1.0 / factor, inv_factor):
                ny, nx = self.arr.shape

[...]

PR #1185, opened 2026-06-23
Add option for flux conserving resample

madelinemarshall:
Fix for Issue #1183, Negative pixels introduced in resampling.
Includes an optional flux_conserving_resample (bool) attribute in the Image class. If switched to True, this implements flux-conserving integer resampling: downsampling is performed by summing pixel blocks, and supersampling is performed by uniformly splitting each pixel into subpixels. This preserves the total flux, but requires integer resampling factors. If flux_conserving_resample=False (default), code uses the original resampling implementation, a spline interpolation.

Trying my best to do the PR on my own, but I don't know what I'm doing so someone should definitely check this. 
"I have added tests that prove my fix is effective or that my feature works" - I did this manually, but I haven't implemented any inbuilt testing.
"New and existing unit tests pass locally with my changes" - I don't know how to do this, but ruff and pre-commit both passed.

## Issue Type

- Enhancement

## Checklist
- [X] I have read the [CONTRIBUTING.md]() -->
- [X] I have added docstrings to all methods
- [X] I have added sufficient comments to all lines
- [X] My changes generate no pep8 errors

## Summary by CodeRabbit

* **Improvements**
  * Image resampling algorithm enhanced with flux-conserving methods for integer-factor and near-integer-factor scaling, improving data preservation during upsampling and downsampling operations.

* **Bug Fixes**
  * Stronger input validation for resampling operations with warnings for edge cases and error handling for invalid scaling factors.

### DIFF
--- a/src/synthesizer/imaging/image.py
+++ b/src/synthesizer/imaging/image.py
@@ -279,12 +279,20 @@ class Image(ImagingBase):
                 self.resolution,
                 self.fov,
                 img=self.arr + other_img.arr,
+                flux_conserving_resample=(
+                    self.flux_conserving_resample
+                    or other_img.flux_conserving_resample
+                ),
             )
         elif self.units is not None and other_img.units is not None:
             return Image(
                 self.resolution,
                 self.fov,
                 img=self.arr * self.units + other_img.arr * other_img.units,
+                flux_conserving_resample=(
+                    self.flux_conserving_resample
+                    or other_img.flux_conserving_resample
+                ),
             )
         else:
             s = "dimensionless"
@@ -330,12 +338,20 @@ class Image(ImagingBase):
                 self.resolution,
                 self.fov,
                 img=self.arr - other_img.arr,
+                flux_conserving_resample=(
+                    self.flux_conserving_resample
+                    or other_img.flux_conserving_resample
+                ),
             )
         elif self.units is not None and other_img.units is not None:
             return Image(
                 self.resolution,
                 self.fov,
                 img=self.arr * self.units - other_img.arr * other_img.units,
+                flux_conserving_resample=(
+                    self.flux_conserving_resample
+                    or other_img.flux_conserving_resample
+                ),
             )
         else:
             s = "dimensionless"
@@ -369,7 +385,7 @@ class Image(ImagingBase):
                 The new image containing the multipled array.
         """
         # Create the new image
-        new_img = Image(self.resolution, self.fov)
+        new_img = Image(self.resolution, self.fov, self.flux_conserving_resample)
 
         # Associate the image array and units
         new_img.arr = self.arr.copy()
--- a/src/synthesizer/imaging/image.py
+++ b/src/synthesizer/imaging/image.py
@@ -62,10 +62,10 @@ class Image(ImagingBase):
             The weight map derived from the noise array.
         flux_conserving_resample (bool):
             If True, use flux-conserving integer resampling: downsampling is
-            performed by summing pixel blocks, and supersampling is performed by
-            uniformly splitting each pixel into subpixels. This preserves the
-            total flux, but requires integer resampling factors. Otherwise, use
-            standard spline interpolation when resampling.
+            performed by summing pixel blocks, and supersampling is performed
+            by uniformly splitting each pixel into subpixels. This preserves
+            the total flux, but requires integer resampling factors.
+            Otherwise, use standard spline interpolation when resampling.
     """
 
     @timed("Image.__init__")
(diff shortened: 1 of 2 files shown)


WillJRoper:
Thanks, @madelinemarshall. The testing infrastructure is all automatic, so you've done the right thing. Adding new tests would be adding unit tests somewhere in the `tests/` directory. Honestly, for this, an AI will do a very good job of introducing some tests if you wanted to do that.

src/synthesizer/imaging/image.py
```diff
@@ -148,7 +164,51 @@ def resample(self, factor):
...
+
+                else:
+                    raise ValueError(
+                        "flux_conserving_resample requires the resampling "
+                        "factor or its reciprocal to be integer, but received "
+                        f"{factor}."
+                    )
+            else:
+                # Resample via spline interpolation
+                self.arr = zoom(self.arr, (factor, factor), order=3)
```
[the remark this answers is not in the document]
WillJRoper: This is quite a subtle issue but I think it's right.

src/synthesizer/imaging/image.py
```diff
@@ -68,6 +74,7 @@ def __init__(
         resolution,
         fov,
         img=None,
+        flux_conserving_resample=False,
```
WillJRoper: Do we need to store this on the `Image`? Is there a reason it needs to know the resampling flag?

WillJRoper: I'd instead have this be an argument to the resample method to dictate which resampling to use.

WillJRoper (changes requested):
Other than the open rabbit comment and the removal of the attribute, this looks good. Maybe I'm missing something but I don't think we should store the flag.

I do wonder long term if there is a one-size-fits-all solution that would be safe for float resample factors, but I also don't think that is a big limitation in the grand scheme.

src/synthesizer/imaging/image.py
madelinemarshall: Hey Will, it's a good question. My logic for doing this on the Image side was that I can pass that in to my standard imaging code when I first create the images, it's an easy user interface. I got the impression that a lot of the e.g. get_image_flux and apply_psf functions that actually call resample happen a bit more behind-the-scenes and so I wasn't exactly sure how to add in a function argument that would get carried all the way through to where the resample gets called. This way it's just attached to the image so I don't have to follow where it goes. Would you like me to try and do it as an argument into resample instead?

WillJRoper: I get the thinking 🤔 ideally, I think I would prefer it getting passed in to `resample`, but you are right that its a bit of a pain to follow. I think this is actually an `Instrument` property, really isn't it? Which is where `apply_psf` and whatnot are defined. I can imagine the user setting up their instrument and wanting to flag there what resampling to use. 

OR the better solution with less of a headache is to only ever use this flux-conserving resampling path? I would honestly be happy with that. Bin the old version, bin the flag, and always enforce integer resampling. I don't think we care about having float resample factors and that sort of change will have a very small ripple effect ot other's scripts. We could even just force an integer with `int` (as you do) and simply put a warning to tell the user if they passed a non-int. That way its a change in behaviour not a breaking change,

WillJRoper: Sorry, very stream of consciousness there... but in short: I prefer the nuclear second option

madelinemarshall: The nuclear second option is certainly a lot easier for me to implement! 

I do think we'd still want a non integer option that goes through the original spline. I imagine there are some cases where the input PSF is on a native instrument pixel scale (e.g. JWST NIRCam 0.031'') while the user wants something else (e.g. 0.02'' or 0.03'' are common for drizzled outputs of real NIRCam images). So I worry if I remove option that then some things might start to break for people!

I don't know enough about the code and how people to use it to really know whether always enforcing this new method (if an integer factor is input) would cause any problems. In my specific use case the integer method works better and is more reliable! But that not be how it's always used by others. But if you're okay with implementing this new method automatically for all, then I'd suggest I do that (remove all the flagging) and keep the original spline in the non-integer case.

WillJRoper: Good point! Let's keep both but as you suggest, switch the method based on whether we have been given an int or a float. 

My only concern is keeping complexity down. That is the easiest solution all round and easy to document that behaviour with minimal changes.

madelinemarshall:
Hey Will, I think I've made all the changes we talked about. The CodeRabbit review suggests to add a warning when the code uses the spline instead of the flux-conserving version, but I think that might be overkill. The user isn't making any choices about which method to use so it's not like they'd be necessarily expecting a specific option to occur (and be confused if the other option occurred instead). Given how many times this call gets used in a set of image generations, it'd just be a lot of output for no great reason. Happy to add this though if you think we should.

WillJRoper (approved):

WillJRoper:
Sorry for the delay in merging this. I've been decidedly MIA. I agree with your reasoning on the warning.

madelinemarshall:
No worries, thanks Will!

PR #1188, opened 2026-07-27
Add SommovigoBartlett2026 attenuation curve and predictive scaling re…

LauraSommovigo:
Add SommovigoBartlett2026 attenuation curve and predictive scaling relations from galaxy properties, based on this recent paper: https://arxiv.org/abs/2606.10027

## Issue Type

- Enhancement

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a unified interface for power-law, Calzetti, Milky Way, grain-based, Li08, Draine & Li, and Sommovigo & Bartlett (2026) attenuation models.
  * Added attenuation, transmission, extinction-curve, and plotting capabilities.
  * Added galaxy-property-based SB26 parameter prediction with optional noise and multiple dust mixtures.
  * Updated examples to visualize SB26 curves and predictions.
* **Bug Fixes**
  * Improved mixed scalar and array handling, validation, and numerical stability for SB26 predictions.
* **Documentation**
  * Corrected attenuation-curve descriptions and references.


src/synthesizer/emission_models/transformers/dust_attenuation.py
```diff
@@ -1181,6 +1183,379 @@ def get_tau_at_lam(self, lam):
...
+                The optical depth.
+        """
+        raise exceptions.UnimplementedFunctionality(
+            "SommovigoBartlett2026 is fit to normalised A_lam/A_V "
+            "values, so get_tau_at_lam is not defined. Use get_tau "
+            "for the normalised curve."
+        )
+
+
+def predict_sommovigo_bartlett_2026(
```
WillJRoper, line 1381: Does this need to be separate from the class? From the example, it just seems to be used to generate the inputs to the class; could it not be folded into the class and become part of the initialisation of the model?

src/synthesizer/emission_models/transformers/dust_attenuation.py
```diff
@@ -1181,6 +1183,379 @@ def get_tau_at_lam(self, lam):
...
+        return SB26Tau(
+            lam=lam,
+            B_0=self.B_0,
+            B_1s=self.B_1s,
+            B_2s=self.B_2s,
+            B_3=self.B_3,
+        )
+
+    @accepts(lam=angstrom)
+    def get_tau_at_lam(self, lam):
```
WillJRoper, line 1362: Maybe I am missing something but can't this method still be implemented like some of the other transformers?

src/synthesizer/emission_models/transformers/dust_attenuation.py
```diff
@@ -1181,6 +1183,379 @@ def get_tau_at_lam(self, lam):
...
+# SommovigoBartlett2026 galaxy-property prediction pipeline.
+_SB26_DUST_FRACTIONS = {
+    "MW": 0.5968,
+    "SMC": 0.7193,
+    "stellar": 0.3064,
+}
+
+
+@accepts(lam=angstrom)
+def SB26Tau(lam, B_0, B_1s, B_2s, B_3):
```
WillJRoper, line 1196: This is the height of style pedantary on my part but functions and variables should not contain uppercase letters, only class names and constants can.

src/synthesizer/emission_models/transformers/dust_attenuation.py
```diff
@@ -50,6 +50,8 @@
     "Calzetti2000",
     "GrainModels",
     "ParametricLi08",
+    "SommovigoBartlett2026",
+    "predict_sommovigo_bartlett_2026",
```
WillJRoper, line 54: As with the other comment, I would rather not expose the predict function as well as the transformer in the public API.

WillJRoper (changes requested):
Thanks for contributing again @LauraSommovigo! The AI reviewer has caught a few things (ranging from pedantic documentation comments to some safety things) and I've got some minor comments including one structural one. 

I may be missing a reason things are structured the way they are so do push back if that is the case.

src/synthesizer/emission_models/transformers/dust_attenuation.py
LauraSommovigo: Hey Will. I am not sure. I wanted in principle to have the functional form be separate from the prediction from physical properties, as one might one to adopt the first but not the latter. Happy to modify if this is not appropriate. I am fixing the rest, sorry i was in applications hell

LauraSommovigo: I'll think of a different structure, based also on the subsequent comment and submit that, you can tell me if that works

WillJRoper: No worries at all! I hope the application hell is fruitful! Been there, done that, now run an absolute mile from it! 😂 

I can see the justification for wanting that separation, and I think you're right that having the separation is good. I would keep the separate function to make this separation clear. 

There is one wrinkle though, and it's a subtle one that I only just realised! If the attenuation curve itself takes the parameters as arguments rather than galaxy properties, then the curve can't dynamically adapt to a whole sample of galaxies. To be clear, this is more of a Synthesizer workflow oddity than an issue in your implementation! I'll try to explain why in a paragraph below but in short, the class needs to take the phyical properties as arguments so it can be signalled to extract galaxy properties at runtime. 

The long reasoning (skip paragrpah and see below if you don't want the nitty gritty), is that in Synthesizer we have a hierarchy for parameters which exists solely to maximise flexibility, variability, and performance. At runtime we will find a parameter a model (or transformer in this case) needs and look for it first on the model (assuming a fixed value has been set for the parameter on the model itself, this is a fixed value will be used whenever the model is evaluated), then if this overide is not found it'll check the "emitter" i.e. galaxy, stars, or black hole object. If the parameter is set to a string (say "tau_v" for an optical depth) then the extraction will recurse and check the model and emitter again but for this new parameter. This means if in your model I want to use each galaxy's `log10_mstar` attribute (which you set yourself on the obejcts) then I can simply pass that string to the appropriate argument. 

This is all a bit complex and is mainly a API into the way Synthesizer works, its probably a sign that I need to improve how this is documented to help people with their implementations of `Transformers`!!!

WillJRoper: I would equally be happy to let this in in its current state as you have written it (modulo the less structural comments from coderabbit and me), and then make these changes myself to fully integrate things (leaving the functionality you have here intact). I can then invite you to review my modifications to make sure you are happy with the edits? 

The last thing I want to do is overcomplicate your workload because this is a very nice addition!

src/synthesizer/emission_models/transformers/dust_attenuation.py
LauraSommovigo: i'll change the function name but the parameters where upper case in the paper to distinguish them as "variables" from symbolic regression vs the numerical constants that were al lower case. might be confusing to switch notation here, but if need be i'll do so

WillJRoper: Know that it pains me to agree, but consistency with the paper makes sense. I'll let this lie 😂

src/synthesizer/emission_models/transformers/dust_attenuation.py
LauraSommovigo: yean not loving applications for sure ... I also realized i had changed my primary email on github but not my user email, so notifications on your replies went to my PhD discontinued email, but all good now i fixed it. SO.. i I've pushed a version where predict is a classmethod on the class, so the separation is there but under one namespace. But I understand now that for full Synthesizer integration the class itself may need to accept galaxy properties (sigma_sfr, inclination, etc.) directly, so the runtime extraction can feed per-galaxy values. Happy to restructure if you can point me to an existing transformer or if you'd prefer to handle that part of the integration yourself given it's tightly coupled to the Synthesizer internals all good i can have a look after. Let me know if anything on my end is unclear

LauraSommovigo: everything else should have been fixed, i will count my legendary win with the parameters upper case

WillJRoper: I'm really conflicted 😅 I've had a look at your changes, and I'm happy as is, but I wonder if we actually want to have the galaxies store the parameters like. We already have precedent for this with `calculate_dust_to_metal_vijayan19`. 

What made me think of this was your required parameters list (which is what the internals will use to discover parameters). If the (uppercase) parameters are what is left as the required parameters, then it makes sense to do it this way and have a method on a galaxy that populates those parameters on the emitter. This also makes a lot of sense since I noticed one of the parameters is the gas metallicity which isn't readily available on a stars object (unless you add it there of course, anything is possible). The downside to this being that every galaxy would need this method called to compute their parameters for your dust model.

Which would you prefer? A method to populate the parameters on a galaxy, or the class itself to take the physical properties as arguments? 

I am more than happy to implement this, as I think it'll make a lot more sense to see in action and then I'll just ask for your review on that PR?

LauraSommovigo: I am not sure tbh, I think either would work? as long as the user has all the info?

WillJRoper: I think, given the need for gas information, the galaxy-level method makes the most sense, but it might be a bit of a "wait and see" where it'll be clear once implemented if that does make sense or not. 

How do you define the gas metallicity? Is it an aperture-smoothed metallicity? Or just the metallicity assigned to the star particles? (I suppose I am thinking about the particle case specifically here, the parametric one is simpler).

WillJRoper: FYI @LauraSommovigo, thought a bit more about this this morning. I will merge this PR as is, and then I'll make some small modifications to integrate it and ask you to be the reviewer on that PR. What I do depends on the answer to the above. 

If gas metallicity is simply the stellar particle metallicity (assuming that that metallicity is representative of the unresolved gas in and around the stellar population) then I will have the class take the physical properties as inputs. 

If it is a true gas metallicity (which would require some sort of definition for particles, i.e. kernel) then I will go for the method version since that method can automate the calculation of required properties. 

Let me know which way makes the most sense, and I will merge and then get to work.

LauraSommovigo: Zgas is the gas-phase metallicity (mass-weighted, from TNG gas cells / GFM_Metallicity in their catalog), not the stellar particle metallicity. I had checked that the change compared to the stellar metallicity depending on aperture radius was not huge but that's that. It's a single scalar per galaxy from the snap_93 catalogs, in absolute mass fraction, that i then rescale to solar metallicity assuming Z_sun = 0.0142

WillJRoper: Cool cool, definitely the method side of things then! 

General question that has no bearing on this implementation: Could this eventually be used to give each star particle its own attenuation curve based on its local properties? You could envision mapping to particle properties rather than integrated and give each particle a unique curve.

WillJRoper (approved):

src/synthesizer/emission_models/transformers/dust_attenuation.py
LauraSommovigo: in principle yes, that would be an extinction curve though, and these functional forms have been derived as attenuation curve for the global galaxy through different los, so not sure how well it will work in this fashion? Sophie N. is doing quite a lot of testing with attenuation curves functional forms applies to SC-SAM + synthesizer, albeit using atm the other Li08 functional form. I want her to try this one too, so she could be the person to have a look at this? ot anyone else, happy to discuss

PR #1202, opened 2026-09-04
Fix mass calculations at earlier times.

jackcturner:
The age binning and clipping logic used by `get_initial_mass_at_age` and `get_surviving_mass_at_age` produced strange behaviour at lookback times close to the max_age of the parametric `Stars`. Rather than naturally tending to zero, it would reach a non-zero floor before dropping to zero instantaneously when age > max_age. 

`get_at_earlier_time` and the aforementioned functions now use a helper, `_get_normalised_sfzh`, which returns the normalised SFZH at a given age_offset. The initial and surviving masses are then computed consistently with initialisation, giving the more physical result shown in the second figure.

constant_orig
constant_fix

## Issue Type

- Bug

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stellar population calculations at earlier ages for both generated and explicitly provided star-formation histories.
  - Corrected surviving-mass normalization to account for stellar evolution fractions consistently.
  - Preserved star-formation history shapes and mass consistency when calculating initial and surviving masses at different ages.
  - Prevented age-based calculations from modifying the original star-formation history data.
  - Improved handling of mixed and array-based star-formation history inputs.


src/synthesizer/parametric/stars.py
```diff
@@ -544,34 +544,50 @@ def _get_sfzh(self, age_offset=None):
...
+        Args:
+            age_offset (unyt_quantity):
+                The offset to apply to the age grid when calculating the SFZH.
+
+        Returns:
+            np.ndarray:
+                The mass-scaled SFZH grid.
+        """
+        # Get the shape of the SFZH.
+        sfzh = self._get_sfzh(age_offset=age_offset)
```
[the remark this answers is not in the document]
WillJRoper: @jackcturner I think this is actually a valid concern. `_get_sfzh` was initially designed as an abstraction of the init process and thus does some weird things because it was written from the standpoint of setting things up the first time. It might need a bit of cleanup in general, but for now this is an issue. Not sure what the simplest solution is, honestly; the AI has suggested something, but its not clear to me that that is the cleanest fix.

src/synthesizer/parametric/stars.py
```diff
@@ -544,34 +544,50 @@ def _get_sfzh(self, age_offset=None):
         # Finally, calculate the SFZH grid based on the above calculations
         return sf_hist[:, np.newaxis] * metal_dist
 
+    def _get_normalised_sfzh(self, age_offset=None):
```
WillJRoper: Minor, but the age_offset=None default here will double-normalise. __init__ overwrites self.sf_hist and self.metal_dist at 343-344 with the already-normalised projections, and _get_sfzh(None) takes the branch that reuses them, so line 566 scales a second time. Nothing hits it today since every caller passes something, but I'd either make the argument required or skip the normalisation when it's None, otherwise someone will trip over it later.

src/synthesizer/parametric/stars.py
```diff
@@ -544,34 +544,50 @@ def _get_sfzh(self, age_offset=None):
...
+        Returns:
+            np.ndarray:
+                The mass-scaled SFZH grid.
+        """
+        # Get the shape of the SFZH.
+        sfzh = self._get_sfzh(age_offset=age_offset)
+
+        # Apply the normalisation to total mass if available.
+        if getattr(self, "sfzh_normalisation", None) is not None:
+            sfzh = sfzh * self.sfzh_normalisation
```
WillJRoper: Flagged this separately in #1203, but it has a huge effect here too — when `sfzh_normalisation` is the 2D array from the surviving mass branch, this multiplies bin (i, j) of the shifted grid by 1/f[i, j], and that f was matched to the unshifted population, so the survival factors end up at the wrong ages. 

Feel free to address this in this PR or we can tackle it separately.

WillJRoper (changes requested):
Couple minor things and the rabbits thoughts. The workflow was also failing, which is related to the rabbits complaint.

src/synthesizer/parametric/stars.py
```diff
@@ -544,34 +545,102 @@ def _get_sfzh(self, age_offset=None):
...
+                The offset to apply to the age grid when calculating the SFZH.
+
+        Returns:
+            sfzh (np.ndarray):
+                The SFZH grid at the earlier lookback time.
+        """
+        # Re-integrate the SFZH over the shifted age bins if it is
+        # function based.
+        if self.sf_hist_func is not None or self.metal_dist_func is not None:
+            sfzh = self._get_sfzh(age_offset=age_offset)
```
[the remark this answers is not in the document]
jackcturner: While the original suggestion was correct and now hopefully addressed, I don't agree with this. If `self.sf_hist_func = None`, then `_get_sfzh` ignores the age_offset completely, returning the original SFH. I think it would be better to simply change this condition to `and`, with a view to tidying up `_get_sfzh` in the future as suggested.

src/synthesizer/parametric/stars.py
jackcturner: I've opted to make it required.

src/synthesizer/parametric/stars.py
jackcturner: I don't know if I've addressed this fully, but I've changed the way the normalisation is applied.

WillJRoper (approved):

